PR #20, opened 2022-09-13
feat: add unpack variants option

mdegat01:
Add a flag to unpack variants in outputs. Supported this flag in all methods of the proxy interface (methods, properties and signals).

Support for this flag was not implemented in `MessageBus.call`. We could do this but it seemed like that one should remain raw. Also the wrappers in Proxy Interface won't work correctly if variants are unpacked by call so they have to handle it separately anyway.

### DIFF
--- a/src/dbus_fast/aio/proxy_object.py
+++ b/src/dbus_fast/aio/proxy_object.py
@@ -8,7 +8,8 @@ from ..errors import DBusError
 from ..message import Message, MessageFlag
 from ..message_bus import BaseMessageBus
 from ..proxy_object import BaseProxyInterface, BaseProxyObject
-from ..signature import Variant, unpack_variants
+from ..signature import Variant
+from ..signature import unpack_variants as unpack
 
 
 class ProxyInterface(BaseProxyInterface):
@@ -74,7 +75,9 @@ class ProxyInterface(BaseProxyInterface):
     """
 
     def _add_method(self, intr_method):
-        async def method_fn(*args, flags=MessageFlag.NONE):
+        async def method_fn(
+            *args, flags=MessageFlag.NONE, unpack_variants: bool = False
+        ):
             input_body, unix_fds = replace_fds_with_idx(
                 intr_method.in_signature, list(args)
             )
@@ -87,7 +90,7 @@ class ProxyInterface(BaseProxyInterface):
                     member=intr_method.name,
                     signature=intr_method.in_signature,
                     body=input_body,
-                    flags=flags & ~MessageFlag.UNPACK_VARIANTS,
+                    flags=flags,
                     unix_fds=unix_fds,
                 )
             )
@@ -104,8 +107,8 @@ class ProxyInterface(BaseProxyInterface):
             if not out_len:
                 return None
 
-            if flags & MessageFlag.UNPACK_VARIANTS:
-                body = unpack_variants(body)
+            if unpack_variants:
+                body = unpack(body)
 
             if out_len == 1:
                 return body[0]
@@ -118,7 +121,9 @@ class ProxyInterface(BaseProxyInterface):
         self,
         intr_property,
     ):
-        async def property_getter(*, flags=MessageFlag.NONE):
+        async def property_getter(
+            *, flags=MessageFlag.NONE, unpack_variants: bool = False
+        ):
             msg = await self.bus.call(
                 Message(
                     destination=self.bus_name,
@@ -141,8 +146,8 @@ class ProxyInterface(BaseProxyInterface):
 
             body = replace_idx_with_fds("v", msg.body, msg.unix_fds)[0].value
 
-            if flags & MessageFlag.UNPACK_VARIANTS:
-                return unpack_variants(body)
+            if unpack_variants:
+                return unpack(body)
             return body
 
         async def property_setter(val):
--- a/src/dbus_fast/aio/proxy_object.py
+++ b/src/dbus_fast/aio/proxy_object.py
@@ -8,7 +8,7 @@ from ..errors import DBusError
 from ..message import Message, MessageFlag
 from ..message_bus import BaseMessageBus
 from ..proxy_object import BaseProxyInterface, BaseProxyObject
-from ..signature import Variant
+from ..signature import Variant, unpack_variants
 
 
 class ProxyInterface(BaseProxyInterface):
@@ -87,11 +87,7 @@ class ProxyInterface(BaseProxyInterface):
                     member=intr_method.name,
                     signature=intr_method.in_signature,
                     body=input_body,
-                    flags=(
-                        flags ^ MessageFlag.REMOVE_SIGNATURE
-                        if flags & MessageFlag.REMOVE_SIGNATURE
-                        else flags
-                    ),
+                    flags=flags & ~MessageFlag.UNPACK_VARIANTS,
                     unix_fds=unix_fds,
                 )
             )
@@ -108,8 +104,8 @@ class ProxyInterface(BaseProxyInterface):
             if not out_len:
                 return None
 
-            if flags & MessageFlag.REMOVE_SIGNATURE:
-                body = BaseProxyInterface.remove_signature(body)
+            if flags & MessageFlag.UNPACK_VARIANTS:
+                body = unpack_variants(body)
 
             if out_len == 1:
                 return body[0]
@@ -145,8 +141,8 @@ class ProxyInterface(BaseProxyInterface):
 
             body = replace_idx_with_fds("v", msg.body, msg.unix_fds)[0].value
 
-            if flags & MessageFlag.REMOVE_SIGNATURE:
-                return BaseProxyInterface.remove_signature(body)
+            if flags & MessageFlag.UNPACK_VARIANTS:
+                return unpack_variants(body)
             return body
 
         async def property_setter(val):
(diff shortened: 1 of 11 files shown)


src/dbus_fast/aio/proxy_object.py
```diff
@@ -87,7 +87,11 @@ async def method_fn(*args, flags=MessageFlag.NONE):
                     member=intr_method.name,
                     signature=intr_method.in_signature,
                     body=input_body,
-                    flags=flags,
+                    flags=(
+                        flags ^ MessageFlag.REMOVE_SIGNATURE
+                        if flags & MessageFlag.REMOVE_SIGNATURE
+                        else flags
+                    ),
```
dlech, line 94:
```suggestion
                    flags=flags & ~MessageFlag.REMOVE_SIGNATURE,
```

dlech: Or instead of silently fixing it, throw a ValueError?

src/dbus_fast/proxy_object.py
```diff
@@ -72,6 +83,22 @@ def _check_method_return(msg, signature=None):
                 msg,
             )
 
+    @staticmethod
+    def remove_signature(data: Any):
```
dlech, line 87: I think it would be better/safer if this was a pure function and didn't modify `data`.

dlech: What is the reason for making this a method on the `BaseProxyInterface` class? It seems like it could just be a standalone function and it doesn't seem to be particularly related to proxies.

src/dbus_fast/proxy_object.py
```diff
@@ -46,7 +57,7 @@ def __init__(self, bus_name, path, introspection, bus):
         self.path = path
         self.introspection = introspection
         self.bus = bus
-        self._signal_handlers = {}
+        self._signal_handlers: dict[str, list[SignalHandler]] = {}
```
dlech, line 60: Need to use `typing` types for compatibility with older Python versions.

```suggestion
        self._signal_handlers: Dict[str, List[SignalHandler]] = {}
```

dlech: this change seems unrelated to the rest of the PR though.

dlech (commented):
This is something we are using in Bleak, so it could be nice to have it built-in to dbus-fast.

Although I could not understand what "remove signatures" meant until I read all of the code. In Bleak we called this "unpack_variants" since that is basically what it is doing. When I read "remove signatures", I thought this was doing some really low-level thing with the d-bus message itself.

src/dbus_fast/aio/proxy_object.py
mdegat01: It doesn't fix it actually. It's totally fine to pass in `REMOVE_SIGNATURE` to this method. The reason its doing this is in case we ever want to support the `REMOVE_SIGNATURE` flag in the underlying `MessageBus.call` method. If the signature is stripped by the call method method then this method (in the proxy object) will break. As the code that follows relies on checking the signature.

Basically I just want to be able to handle removing the variants/signature data in this code and not have `call` ever strip it out.

mdegat01: Also that is a way better way to do this, will change it 👍

src/dbus_fast/proxy_object.py
mdegat01: So in truth I was just moving it from here:
https://github.com/home-assistant/supervisor/blob/b8249548aef05be3353fb6dc45b7df4a200bfa53/supervisor/utils/dbus.py#L56-L71

I'm fine with making it a standalone function. As for making it functional and not modify data though, that's a trickier. It absolutely makes sense but that means making a deep copy and @bdraco and I were worried about the performance hit there.

src/dbus_fast/proxy_object.py
mdegat01: > this change seems unrelated to the rest of the PR though.

Yes and no. I did it because this PR changed the type of `_signal_handlers`. It was previously type `dict[str, list[Callable]]` and after this PR it will be `dict[str, list[SignalHandler]]`. That was the reason for it, I wanted to make it clear what type it is now and that it had changed. But I am fine with dropping it if we want.

Also I did forget about the 3.7 rules, thanks for the reminder.

mdegat01:
Unpack variants is a good name @dlech . As I mentioned above I was really just moving this code from supervisor into the library as it seemed really useful for everyone and kept the names we were using the same. But I'll rename it, that does make more sense.

src/dbus_fast/proxy_object.py
bdraco: If there is any kind of volume we should avoid the deepcopy. If its not high volume it won't matter

dlech: The function already makes a copy of lists and reassigns every key in dictionaries, so it seems like it is already doing as much work as a deep copy. And there is a separate `deepcopy` in one of the calls I saw, so in that case, it would be doing the equivalent work of two deep copies which could be reduced to 1 if this was a pure function.

bdraco: Will generate a profile to see how much its costing in bleak

bdraco: Its about 4.1% of the run time of processing incoming messages in bleak.  The `isinstance` checks and self calling look like they are more expensive than the actual copying

unpack_variants

mdegat01: Fair enough. I'll make the change then.

bdraco: https://github.com/hbldh/bleak/blob/d343cdd96cf6eb3de19d1dc6a85af664a96bc36d/bleak/backends/bluezdbus/utils.py#L30

mdegat01: The added deepcopy was only for signals for reference. That was mainly because I noticed that if you have multiple signal handlers for the same signal and one of the handlers does modify data it actually affects the other handler. Since the same object is passed into each. So in that one case I knew I had to make a deepcopy to remove the variants otherwise it would be impossible to have one signal handler that wanted the raw response and another that didn't.

Although as an aside we may want to address that in a future PR. I don't think one signal handler should be able to affect the others.

mdegat01:
So for the codecov check, its basically failing that because I also modified the `glib` part for completeness. And I also added tests for that but I see those aren't running. Should I just remove all changes to the `glib` and only change `aio`?

src/dbus_fast/proxy_object.py
mdegat01: FYI this function is moved to just be a function in `signature.py`. Left this convo unresolved because it seemed like it was still under discussion around the performance bits though

bdraco:
> So for the codecov check, its basically failing that because I also modified the `glib` part for completeness. And I also added tests for that but I see those aren't running. Should I just remove all changes to the `glib` and only change `aio`?

I think we are missing a library being installed in the CI.  We should probably fix that first.  I'll see if I can fix it before I go to sleep but getting late here so not if I can before sleep.

bdraco:
GI tests should work now. CI should be a bit faster as well as caching will be turned on with this PR

https://github.com/Bluetooth-Devices/dbus-fast/pull/21

I'm too tired to be sure I got everything right with the jet lag so I didn't merge it, but feel free if it looks ok.

g'night

src/dbus_fast/aio/proxy_object.py
```diff
@@ -103,16 +103,22 @@ async def method_fn(*args, flags=MessageFlag.NONE):
...
         setattr(self, method_name, method_fn)
 
-    def _add_property(self, intr_property):
-        async def property_getter():
+    def _add_property(
+        self,
+        intr_property,
+    ):
+        async def property_getter(*, flags=MessageFlag.NONE):
             msg = await self.bus.call(
```
bdraco, line 122: In a future PR it would be good to move the unpack variants as part of `call` as well so we don't create them in the first place

mdegat01: Ok yea that makes sense. I wasn't sure if we wanted it in the low-level client but I knew even if we did the high-level client needed the variants returned for now because otherwise this bit would fail:
https://github.com/Bluetooth-Devices/dbus-fast/blob/f4d173e0426990cd0b8b1b813949cd5027684102/src/dbus_fast/aio/proxy_object.py#L98-L103

We'd have to refactor that out as well if the high level client is not being returned messages that contain variants.

bdraco: It is a bit difficult since everything is blindly unpacked.  Anyways, something for the future if we can figure out a workable solution

bdraco (approved):

src/dbus_fast/constants.py
```diff
@@ -29,6 +29,7 @@ class MessageFlag(IntFlag):
     NO_REPLY_EXPECTED = 1  #: The method call does not expect a method return.
     NO_AUTOSTART = 2
     ALLOW_INTERACTIVE_AUTHORIZATION = 4
+    UNPACK_VARIANTS = 8
```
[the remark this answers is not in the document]
mdegat01: Oh! I didn't realize that, thanks for letting me know. Ok in that case I won't use flags at all and just switch to a new keyword arg.

bdraco (approved):
LGTM 👍

bdraco (approved):

PR #21, opened 2022-09-14
chore: run gi tests in the CI

bdraco:


.github/workflows/ci.yml
```diff
@@ -46,14 +46,19 @@ jobs:
     steps:
       - uses: actions/checkout@v3
       - name: Install libs
-        run: sudo apt-get install -y dbus dbus-x11 python-gi
```
dlech, line 49: This should probably just be `python3-gi`. `python-gi` would install the Python 2 version.

bdraco: Let's give that a shot 👍

bdraco: Looks like they still get skipped since its not for the right python version

bdraco:
Still looks good this morning 😄

PR #22, opened 2022-09-15
feat: allow varargs callback for signals

mdegat01:
The high level client currently does not support callbacks for signals with a varargs parameter. This adds support for that.


src/dbus_fast/proxy_object.py
```diff
@@ -118,8 +118,10 @@ def _message_handler(self, msg):
     def _add_signal(self, intr_signal, interface):
         def on_signal_fn(fn):
             fn_signature = inspect.signature(fn)
-            if not callable(fn) or len(fn_signature.parameters) != len(
```
mdegat01, line 121: Just wanted to note that remove of `not callable(fn)` from here was intentional and is backwards compatible. That was never called. If `fn` was not a callable then `inspect.signature` already raised a `TypeError`. I added a test confirm this behavior below.

bdraco (approved):

PR #25, opened 2022-09-20
fix: disconnect connected buses at end of tests

mdegat01:
Last build failed for confusing reasons. I suspect from the error the real reason is something not being released between tests. Adding a call to `bus.disconnect` at the end of every test which connects to the bus seems to fix it.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -57,7 +57,7 @@ jobs:
       - name: Install Dependencies
         run: poetry install
       - name: Test with Pytest
-        run: export $(dbus-launch); poetry run pytest --cov-report=xml -vvv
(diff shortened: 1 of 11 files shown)


bdraco (approved):

mdegat01:
This change also seems to have dropped the number of warnings in the pytest log about unexpected disconnects and unclosed resources from ~28 to ~7 (depending on the version). Fwiw there was only one test that connected a bus which I did not change and that was this one:
https://github.com/Bluetooth-Devices/dbus-fast/blob/cfad28bd2ba8dccf4c3a591461bb666871e4cbba/tests/test_tcp_address.py#L12-L71

It seemed to be testing what happened when the socket was closed on the bus so I didn't want to change that one around.

PR #29, opened 2022-09-24
feat: improve asyncio write performance

### LINKED ISSUE #28: Writer is added and removed each request
src/dbus_fast/aio/message_bus.py:            self._loop.remove_reader(self._fd)
src/dbus_fast/aio/message_bus.py:            self._loop.remove_writer(self._fd)
        self._loop.add_reader(self._fd, self._message_reader)
Screen Shot 2022-09-23 at 4 39 09 PM
Screen Shot 2022-09-23 at 4 38 46 PM

bdraco:
Similar to https://github.com/python/cpython/blob/2a50772b63d7bc4ef97d16e9bcc53455c301b0ea/Lib/asyncio/selector_events.py?rgh-link-date=2022-09-24T02%3A51%3A37Z#L914

With this change ~99% of writes don't need to be scheduled and can happen right away so it helps quite a bit with the latency

I was able to get sub 1s connection times on my yale locks with this change
```
2022-09-23 21:25:16.427 DEBUG (MainThread) [bleak_retry_connector] Family Room Door (M1028EZ) - /org/bluez/hci3/dev_78_9C_85_08_D2_CE: Connecting (attempt: 1, last rssi: -76)
2022-09-23 21:25:16.727 DEBUG (MainThread) [bleak_retry_connector] Family Room Door (M1028EZ) - /org/bluez/hci3/dev_78_9C_85_08_D2_CE: Connected (attempt: 1, last rssi: -76)
```

Previously the best I could do was slightly > 1s

fixes #28

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -57,7 +57,7 @@ jobs:
       - name: Install Dependencies
         run: poetry install
       - name: Test with Pytest
-        run: export $(dbus-launch); poetry run pytest -vvvs --cov-report=xml
+        run: export $(dbus-launch); poetry run pytest --cov-report=xml
       - name: Upload coverage to Codecov
         uses: codecov/codecov-action@v3
         with:
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -57,7 +57,7 @@ jobs:
       - name: Install Dependencies
         run: poetry install
       - name: Test with Pytest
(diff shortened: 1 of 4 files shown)


bdraco:
I don't think we can get rid of add_writer/remove_writer because its needed for unexpected disconnects.

However with this change ~85% of writes don't need to be scheduled and can happen right away so it helps quite a bit with the latency

PR #30, opened 2022-09-24
feat: improve writer performance with a deque

bdraco:
- The only calls are get_nowait, put_nowait, and empty which means deque is a better fit since we don't need thread safety here


bdraco:
Actually collections.deque might make more sense

PR #31, opened 2022-09-24
feat: add support and workarounds for cpython3.11

bdraco:
Workaround an intentional behavior change in cpython 3.11 https://github.com/python/cpython/issues/98976 by writing out the `MESSAGE_FLAG_MAP`


bdraco:
Seems to hang on `tests/test_aio_low_level.py::test_sending_messages_between_buses`

bdraco:
I think this is actually a bad test that is affecting the other tests but only shows up on py3.11 since it raises sooner

src/dbus_fast/_private/unmarshaller.py
```diff
@@ -590,6 +590,10 @@ def _read_body(self) -> None:
             tree = get_signature_tree(signature)
             body = [self._readers[t.token](self, t) for t in tree.types]
 
+        import pprint
+
+        pprint.pprint([MESSAGE_FLAG_MAP, "flag", self._flag])
```
bdraco, line 595: remove before merging

bdraco:
I think this might be a bug in python 3.11

```
[{1: <MessageFlag.NO_REPLY_EXPECTED: 1>,
  2: <MessageFlag.NO_AUTOSTART: 2>,
  4: <MessageFlag.ALLOW_INTERACTIVE_AUTHORIZATION: 4>},
 'flag',
 0]
```

bdraco:
This is a bug in python3.11

```
Hide these hints with HOMEBREW_NO_ENV_HINTS (see `man brew`).
bdraco@JNK-M1-Pro-New dbus-fast % python3.11 x.py 
  File "/Users/bdraco/dbus-fast/x.py", line 16, in <module>
    assert 0 in MESSAGE_FLAG_MAP, "0 is missing from the map"
AssertionError: 0 is missing from the map
bdraco@JNK-M1-Pro-New dbus-fast % python3.10 x.py 
bdraco@JNK-M1-Pro-New dbus-fast % cat x.py
import pprint
from enum import Enum, IntFlag


class MessageFlag(IntFlag):
    """Flags that affect the behavior of sent and received messages"""

    NONE = 0
    NO_REPLY_EXPECTED = 1  #: The method call does not expect a method return.
    NO_AUTOSTART = 2
    ALLOW_INTERACTIVE_AUTHORIZATION = 4


MESSAGE_FLAG_MAP = {field.value: field for field in MessageFlag}

assert 0 in MESSAGE_FLAG_MAP, "0 is missing from the map"
```

bdraco:
https://github.com/cython/cython/issues/5109

bdraco:
I thought it was cython at first but it fails without it as well

bdraco:
https://github.com/python/cpython/issues/98976

src/dbus_fast/constants.py
```diff
@@ -19,7 +19,13 @@ class MessageType(Enum):
     SIGNAL = 4  #: A broadcast signal to subscribed connections
 
 
-MESSAGE_TYPE_MAP = {field.value: field for field in MessageType}
+# This is written out because of https://github.com/python/cpython/issues/98976
```
ethanfurman, line 22: Only flags are effected by the change in [cpython #98976](https://github.com/python/cpython/issues/98976) .  The change to `MESSAGE_TYPE_MAP` is unnecessary.

ethanfurman: What are the `MESSAGE_*_MAP` dicts used for?

bdraco: Thanks

PR #32, opened 2022-09-24
feat: speed up marshalling

bdraco:


bdraco:
I'm not sure this is worth it as the speed up is only 9%

bdraco:
It actually hits more on bluetooth messages so more like 15% so worth it

PR #38, opened 2022-09-26
feat: use async_timeout instead of asyncio.wait_for

bdraco:
Avoids creating a task which will make it easier
to trace down races in the aio message_bus write_callback

We currently have a race (maybe destruction race) in
https://github.com/Bluetooth-Devices/dbus-fast/blob/d71f0e1fbd7282cbd78402414158545ba7176f6e/src/dbus_fast/aio/message_bus.py#L78

where somehow `self.buf` is `None`

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -31,6 +31,7 @@ sphinxcontrib-fulltoc = {version = "^1.2.0", extras = ["docs"]}
 Sphinx = {version = "^5.1.1", extras = ["docs"]}
 myst-parser = {version = "^0.18.0", extras = ["docs"]}
 sphinx-rtd-theme = {version = "^1.0.0", extras = ["docs"]}
+async-timeout = ">=4.0.1"
 
 [tool.poetry.extras]
 docs = [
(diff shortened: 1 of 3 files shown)


mdegat01 (approved):

pyproject.toml
```diff
@@ -31,6 +31,7 @@ sphinxcontrib-fulltoc = {version = "^1.2.0", extras = ["docs"]}
 Sphinx = {version = "^5.1.1", extras = ["docs"]}
 myst-parser = {version = "^0.18.0", extras = ["docs"]}
 sphinx-rtd-theme = {version = "^1.0.0", extras = ["docs"]}
+async-timeout = ">=4.0.1"
```
dlech, line 34: We had a request in Bleak to also allow v3.x.

https://github.com/hbldh/bleak/pull/1009

so requiring >=4 would cause a regression in Bleak

bdraco: We don't need the 4.x api specifically so we can reduce the pin to 3.x

bdraco: https://github.com/Bluetooth-Devices/dbus-fast/pull/39

We pin it higher in hass to solve dep chain compat with aiohttp but that isn't a problem for `.timeout()` in `async` context so  no worries about loosening the pin here.

PR #39, opened 2022-09-26
fix: loosen async-timeout pin to 3.0.0

bdraco:
https://github.com/Bluetooth-Devices/dbus-fast/pull/38#discussion_r980363091


dlech (approved):
thanks

PR #44, opened 2022-10-01
feat: add optional cython extension

bdraco:
- Improves performance by roughly 60%
- setup of the extension uses the recommended method in https://github.com/python-poetry/poetry/issues/2330 to degrade to pure python gracefully when it can't be compiled to ensure we keep the zero dep and pure Python 3 requirements


bdraco:
The test need to be run with and without the C extension

setup.py
```diff
@@ -1,9 +1,34 @@
...
... [Bluetooth Adapters](https://github.com/bluetooth-devices/bluetooth-adapters)\n\n## Contributing\n\nContributions are welcome. Development happens on [Github](https://github.com/altdesktop/python-dbus-fast).\n\nBefore you commit, run `pre-commit run --all-files` to run the linter, code formatter, and the test suite.\n\n## Copyright\n\nYou can use this code under an MIT license (see LICENSE).\n\n- © 2019, Tony Crisci\n- © 2022, Bluetooth Devices authors\n\n## Contributors ✨\n\nThanks goes to these wonderful people ([emoji key](https://allcontributors.org/docs/en/emoji-key)):\n\n<!-- prettier-ignore-start -->\n<!-- ALL-CONTRIBUTORS-LIST:START - Do not remove or modify this section -->\n<!-- markdownlint-disable -->\n<!-- markdownlint-enable -->\n<!-- ALL-CONTRIBUTORS-LIST:END -->\n<!-- prettier-ignore-end -->\n\nThis project follows the [all-contributors](https://github.com/all-contributors/all-contributors) specification. Contributions of any kind welcome!\n\n## Credits\n\nThis package was created with\n[Cookiecutter](https://github.com/audreyr/cookiecutter) and the\n[browniebroke/cookiecutter-pypackage](https://github.com/browniebroke/cookiecutter-pypackage)\nproject template.\n",
```
balloob, line 17: Does this all really go in there ?

bdraco: It's auto generated by poetry

bdraco: Let me see if I can hack it out without having to maintain it 😉

bdraco: I was actually able to delete it since its not required anymore

README.md
```diff
@@ -48,7 +48,8 @@ Desktop users can use this library to create their own scripts and utilities to
 
 dbus-fast plans to improve over other DBus libraries for Python in the following ways:
 
-- Zero dependencies and pure Python 3.
+- Zero dependencies and pure Python 3
```
dlech, line 51: `asyncio_timeout` is a dependency, so we can no longer claim zero dependencies.

PR #45, opened 2022-10-02
feat: add additional cython types to the unmarshaller

bdraco:
Roughly 16% speedup


src/dbus_fast/_private/unmarshaller.pxd
```diff
@@ -31,9 +31,15 @@ cdef class Unmarshaller:
     @cython.locals(
         start_len=cython.ulong,
         missing_bytes=cython.ulong,
+        data=cython.bytes
```
bdraco, line 34: Likely failing on this since it can be `None`

PR #72, opened 2022-10-04
feat: add cython extension for signature

bdraco:
This is roughly at 10% improvement to the unmarshall path

### DIFF
--- a/src/dbus_fast/message.py
+++ b/src/dbus_fast/message.py
@@ -4,7 +4,7 @@ from ._private.constants import LITTLE_ENDIAN, PROTOCOL_VERSION, HeaderField
 from ._private.marshaller import Marshaller
 from .constants import ErrorType, MessageFlag, MessageType
 from .errors import InvalidMessageError
-from .signature import SignatureTree, Variant
+from .signature import SignatureTree, Variant, get_signature_tree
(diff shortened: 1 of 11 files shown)


.github/workflows/ci.yml
```diff
@@ -65,7 +65,7 @@ jobs:
             poetry install
           fi
       - name: Test with Pytest
-        run: export $(dbus-launch); poetry run pytest --cov-report=xml
+        run: export $(dbus-launch); poetry run pytest -cov-report=xml
```
bdraco, line 68:
```suggestion
        run: export $(dbus-launch); poetry run pytest --cov-report=xml
```

bdraco:
Additional significant improvements would require getting rid of the dict dispatch so we can make the read functions cdef instead of cpdef. That trade off may not be worth it though since it would make the pure python code paths slower.

src/dbus_fast/message.py
```diff
@@ -255,13 +255,6 @@ def new_signal(
             unix_fds=unix_fds,
         )
 
-    def _matches(self, **kwargs):
-        for attr, val in kwargs.items():
```
bdraco, line 259: This is dead code

PR #77, opened 2022-10-06
feat: use SocketIO instead of BufferedRWPair

bdraco:
Performance is suffering from buffering because of all the python copy.  Since we know the size of the message we can avoid reading into a buffer and only reading the size we need

This will do 2 more syscalls but less internal buffering so its up to 50% faster since the syscalls should are far faster than the python code.

### DIFF
--- a/src/dbus_fast/_private/unmarshaller.py
+++ b/src/dbus_fast/_private/unmarshaller.py
@@ -146,7 +146,7 @@ class Unmarshaller:
         "_uint32_unpack",
     )
 
-    def __init__(self, stream: io.BufferedRWPair, sock=None):
+    def __init__(self, stream: socket.SocketIO, sock=None):
         self._unix_fds: List[int] = []
         self._buf = bytearray()  # Actual buffer
         self._view = None  # Memory view of the buffer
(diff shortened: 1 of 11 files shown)


bdraco:
This is a 50% speed up since its no longer doing all the copy

bdraco:
```
[pid    61] recvfrom(25, "l\4\1\0014\0\0\0\322s!\0\225\0\0\0", 16, 0, NULL, NULL) = 16
[pid    61] recvfrom(25, "\1\1o\0%\0\0\0/org/bluez/hci0/dev_60_5"..., 204, 0, NULL, NULL) = 204
```

Two recvfrom per message but all the copy is done at a layer above which avoids all the python string overhead

bdraco:
This actually works great.  The problem is the disconnect is not awaited so its so fast now that its still running and fires later

bdraco:
I'm not sure the speed up I'm seeing is real.

I think the profiler is lying.

bdraco:
pyio can't be this inefficent

bdraco:
The profiler is wrong here.

PR #78, opened 2022-10-06
feat: increase receive buffer size to handle rssi storms

bdraco:


bdraco:
Doesn't matter because the RPI still can't deliver the other side fast enough

PR #82, opened 2022-10-07
feat: unpack header names as message kwargs

bdraco:
speeds up unmarshall by ~8%

### DIFF
--- a/src/dbus_fast/_private/unmarshaller.py
+++ b/src/dbus_fast/_private/unmarshaller.py
@@ -9,13 +9,7 @@ from ..constants import MESSAGE_FLAG_MAP, MESSAGE_TYPE_MAP
 from ..errors import InvalidMessageError
 from ..message import Message
 from ..signature import SignatureType, Variant, get_signature_tree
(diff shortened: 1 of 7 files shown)


src/dbus_fast/_private/unmarshaller.py
```diff
@@ -54,14 +48,19 @@
...
+HEADER_MESSAGE_ARG_NAME = {
+    1: "path",
+    2: "interface",
+    3: "member",
+    4: "error_name",
+    5: "reply_serial",
+    6: "destination",
+    7: "sender",
+    8: "signature",
+    9: "unix_fds",
```
bdraco, line 61: These all look right, but need to be double checked

bdraco: PATH = 1
    INTERFACE = 2
    MEMBER = 3
    ERROR_NAME = 4
    REPLY_SERIAL = 5
    DESTINATION = 6
    SENDER = 7
    SIGNATURE = 8
    UNIX_FDS = 9

PR #99, opened 2022-10-10
feat: additional typing

bdraco:


bdraco:
Right 3.7/3.8 is too old to support subscriptable asyncio.Futures

PR #100, opened 2022-10-10
feat: add additional typing

bdraco:


src/dbus_fast/message.py
```diff
@@ -115,9 +115,7 @@ def __init__(
         self.interface = interface
         self.member = member
         self.message_type = message_type
-        self.flags = (
-            flags if type(flags) is MessageFlag else MessageFlag(bytes([flags]))
-        )
+        self.flags = flags if type(flags) is MessageFlag else MessageFlag(flags)
```
bdraco, line 118: This looks like it was actually a bug

PR #104, opened 2022-10-12
feat: improve aio message reader performance

bdraco:
We can't cythonize `message_bus.py` but its really the reader loop that can use the speed up so break that out into `message_reader.py`

This reduces the read latency by ~17% when the volume is high

### DIFF
--- a/src/dbus_fast/aio/message_reader.py
+++ b/src/dbus_fast/aio/message_reader.py
@@ -14,7 +14,7 @@ def message_reader(
     """Reads messages from the unmarshaller and passes them to the process function."""
     try:
         while True:
-            message = unmarshaller._unmarshall()
+            message = unmarshaller.unmarshall()
             if message:
                 try:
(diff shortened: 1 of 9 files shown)


src/dbus_fast/aio/message_reader.py
```diff
@@ -0,0 +1,35 @@
...
+def build_message_reader(
+    unmarshaller: Unmarshaller,
+    process: Callable[[Message], None],
+    finalize: Callable[[Optional[Exception]], None],
+) -> None:
+    """Build a callable that reads messages from the unmarshaller and passes them to the process function."""
+    unmarshall = unmarshaller.unmarshall
+    reset = unmarshaller.reset
+
+    def _message_reader() -> None:
```
bdraco, line 18: The closure is faster than the bound object and we speed it up even more by pre-unpacking unmarshall and reset since these can get called 20000 times per minute

PR #111, opened 2022-10-20
feat: declare cdefs that do not return as voids

bdraco:


bdraco:
No idea why this stalls forever but runs fine in testing

PR #112, opened 2022-10-23
feat: add typing to dbus_fast._private.util

bdraco:


bdraco:
This one has gone stale. I needs to be reworked. No sense in keeping it open for now

PR #117, opened 2022-10-28
feat: small speed up to unpack_variants

bdraco:


bdraco:
Tested in production 👍

PR #118, opened 2022-10-28
feat: add additional pxd defs for message

bdraco:


bdraco:
Tested in production 👍

PR #127, opened 2022-10-29
fix: pass return value to SendReply.__exit__

dlech:
To return an error in a ServiceInterface, you raise a DBusError which is caught and translated to a message and sent over the bus. However, the context manager that catches this exception was incorrectly not sending a return value of True when the exception was handled. This caused the _message_reader() returned by build_message_reader() to log the error even though it was handled.

### DIFF
--- a/src/dbus_fast/message_bus.py
+++ b/src/dbus_fast/message_bus.py
@@ -762,8 +762,8 @@ class BaseMessageBus:
                 exc_type: Optional[Type[Exception]],
                 exc_value: Optional[Exception],
                 tb: Optional[TracebackType],
-            ) -> None:
-                self._exit(exc_type, exc_value, tb)
+            ) -> bool:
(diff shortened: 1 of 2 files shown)


src/dbus_fast/message_bus.py
```diff
@@ -763,7 +763,7 @@ def __exit__(
                 exc_value: Optional[Exception],
                 tb: Optional[TracebackType],
             ) -> None:
```
bdraco, line 765: Looks like the typing needs to be adjusted as well

dlech: fixed

bdraco (approved):
Thanks!

src/dbus_fast/message_bus.py
bdraco: I've been slowing adding the missing typing when I have cycles. Its about 70% done and then I should be able to turn on mypy in the CI

PR #134, opened 2022-11-01
fix: re-release due to pypi not seeing the new version

bdraco:
https://pypi.org/project/dbus-fast/ shows the new version but the index shows the old version.

https://status.python.org/ shows now issues so trying to do another release to see if it resolves


bdraco:
Looks like semantic-release choked on 3.11 in the title of the previous pr

PR #137, opened 2022-11-02
feat: speed up marshall align

bdraco:
This is a 5% speed up.

By writing this as a loop cython will write
__Pyx_PyByteArray_Append instead of a ~20
lines of python call stack


bdraco:
Tested in production 👍

PR #138, opened 2022-11-02
feat: speed up marshall write_string and write_variant

bdraco:
This is a 7% speed up.

Added some additional cython typing


bdraco:
Tested in production 👍

PR #139, opened 2022-11-02
feat: speed up marshaller by pre-packing bools

bdraco:


bdraco:
Tested in production 👍

PR #142, opened 2022-11-03
feat: speed up marshalling headers

### LINKED ISSUE #140: Marshall optimizations
_write_single can be lined a few more places to reduce code

We can in-line a few more writer as well

We could even support native packing for uint32

bdraco:
This is a 22% speed up

closes #140


bdraco:
Tested in production 👍

PR #143, opened 2022-11-03
feat: add cython def for unmarshaller read_sock for fd passing

### LINKED ISSUE #141: Unmarshaller is not optimized for Unix fds
Reading the sock is much slower. We can probably pre build at least the array struct

bdraco:
closes #141


bdraco:
Tested in production 👍

PR #144, opened 2022-11-03
feat: speed up creating Variant objects

bdraco:
This is only a 1.5% speed up but we create a lot of these


bdraco:
Tested in production 👍

PR #145, opened 2022-11-03
feat: optimize unmarshaller by dropping exception that was only used internally

bdraco:


bdraco:
Tested in production 👍

PR #146, opened 2022-11-04
feat: use cimports for unmarshaller Variant and Message

bdraco:
speed up of 2%


bdraco:
Tested in production 👍

PR #147, opened 2022-11-04
feat: refactor message_reader to avoid python wrappers

bdraco:


bdraco:
With this last change we are down to almost no inner python and just converting data back to pyobjects at the integration points.

PR #148, opened 2022-11-04
feat: small speed up to _unpack_variants

bdraco:
This is a 12-14% speed up


bdraco:
Tested in production 👍

PR #149, opened 2022-11-04
feat: use cimports for message marshalling

bdraco:
This is a 10% speed up


bdraco:
Tested in production 👍

PR #152, opened 2022-11-04
feat: add optimized reader for GetManagedObjects

bdraco:
This is a 4% speed up


bdraco:
Tested on production 👍

PR #159, opened 2022-11-09
feat: use SignatureType cdef for arguments in unmarshaller

bdraco:
This is a 7% speed up


bdraco:
GetManagedObjects is slower with this change, but RSSI is faster

PR #160, opened 2022-11-09
feat: reduce a try block when reading stream

bdraco:


bdraco:
cython makes the try so cheap it doesn't matter

PR #162, opened 2022-11-13
feat: optimize unpack_variants for Variants frequently being dict values

bdraco:


bdraco:
This won't work because we can actually have a variant inside another one

PR #163, opened 2022-11-14
fix: revert str typing for marshaller

bdraco:


mdegat01 (approved):
Yep looks good. Thanks for fixing this, sorry about the hassle. If we want to make this change anyway in a follow-up marked as a breaking change to regain the lost performance I'm fine with it. I'll make sure to get supervisor compliant here anyway.

PR #170, opened 2022-12-01
feat: add return value to method decorator

tom-cook-veea:
This is a very small change to make methods decorated with `@method()` still return their return values in a local context.  This makes it possible to still use a DBus method locally without having to jump through hoops to obtain the return value.  In my case, I often find myself writing code like this:

```
@method()
def PublicFunction() -> 's':
    return do_PublicFunction()

def do_PublicFunction() -> str:
    return "value"
```

so that I have a way of calling the method locally and obtaining the return value.  Since all the decorator does is add a layer of indirection with an extra `__DBUS_METHOD` attribute on the wrapper, there is no problem with calling it locally if that is desired.

### DIFF
--- a/src/dbus_fast/service.py
+++ b/src/dbus_fast/service.py
@@ -94,7 +94,7 @@ def method(name: str = None, disabled: bool = False):
     def decorator(fn):
         @wraps(fn)
         def wrapped(*args, **kwargs):
-            fn(*args, **kwargs)
+            return fn(*args, **kwargs)
 
         fn_name = name if name else fn.__name__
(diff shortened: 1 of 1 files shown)


tom-cook-veea:
PS I'm not really sure why the coverage check has failed here; the relevant code should be hit repeatedly in the existing tests.

bdraco:
Can you add a test that it actually returns the value now?

bdraco:
I'm closing this PR since it has gone stale without a response. https://github.com/Bluetooth-Devices/dbus-fast/pull/170#issuecomment-1340161200

PR #171, opened 2022-12-01
Add message info object

tom-cook-veea:
This change adds a context-local variable that exposes the message information in any message handling context.  The new context-local proxy, `current_message`, can be used as though it was the actual `dbus_fast.message.Message` object.  The context-local has its value set at the start of `_process_message`.

I wanted to discuss this change as it's one of the many ways that this information could be exposed to services; but the change is fairly self-contained and easy to understand so the best way of showing what I'm proposing the change is to just raise the PR.  Feedback welcome.

### DIFF
--- a/src/dbus_fast/message_bus.py
+++ b/src/dbus_fast/message_bus.py
@@ -38,6 +38,7 @@ class ReadOnlyContextProxy:
     immutable.
     :param name: The name of the context variable.
     """
+
     def __init__(self, name: str):
         self._obj = contextvars.ContextVar(name)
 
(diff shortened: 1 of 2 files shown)


tom-cook-veea:
Apologies for getting my commit message formats wrong.  If someone can suggest how I can fix them on an open PR, I am more than willing to do so, but AFAICT it is either not possible or a Very Bad Idea to edit the commit messages of already-pushed commits.

bdraco:
I'm closing this PR since the other one has gone stale without a response. https://github.com/Bluetooth-Devices/dbus-fast/pull/170#issuecomment-1340161200

PR #175, opened 2022-12-08
feat: only construct handlers once

bdraco:
When bleak is using passive scanning each message would cause a method handler to be constructed via `_make_method_handler`

Only construct them once.


bdraco:
Tested in production with passive scanning.

Much better.

There are still parts that search the whole message for fds and will never do anything.

PR #184, opened 2022-12-09
feat: increase socket receive buffer size to reduce pressure on the bus

### LINKED ISSUE #182: Incoming socket buffer is the too small
**Describe the bug**

**To Reproduce**

**Additional context**

bdraco:
closes #182


bdraco:
Doesn't seem to make much difference

Issue #188, opened 2022-12-23
Allow passing hardcoded uid to AuthExternal


kiwiz:
**Is your feature request related to a problem? Please describe.**

I'm trying to connect to the system DBus within a Docker container that is utilizing a user namespace. This fails, as the `os.uid()` call [here](https://github.com/Bluetooth-Devices/dbus-fast/blob/main/src/dbus_fast/auth.py#L69) returns the remapped UID (which is entirely expected).

**Describe the solution you'd like**

For my use case, I would like to pass a hardcoded UID to `AuthExternal`, perhaps via the constructor or an env variable. If this sounds like a reasonable change, I can submit a PR.

**Additional context**

I arrived here via the following path:

* [home-assistant:bluetooth](https://github.com/home-assistant/core/blob/dev/homeassistant/components/bluetooth)
* [bleak](https://github.com/hbldh/bleak)
* dbus-fast

Let me know if this feature request would be better suited for one of those repos.

bdraco:
We can accept a PR for this via the constructor as an optional keyword arg as long as it has test coverage 👍

PR #193, opened 2023-01-04
feat: add support for EXTERNAL auth without uid

mvn23:
As a follow-up to #188 , this PR allows `EXTERNAL` auth without a uid as per RFC4422. If we pass `-1` as the uid, it will trigger the following auth sequence (source: https://gitlab.freedesktop.org/dbus/dbus/-/issues/195):
```
C: AUTH EXTERNAL            # I want to use EXTERNAL, but I do not
                            # have an initial response
S: DATA                     # Who do you claim you are?
C: DATA                     # Whoever the kernel says I am
S: OK                       # Can't argue with that!
```
The upside of this is that when running in a user namespace (e.g. rootless docker), we don't need to know our uid mapping outside of the namespace to connect to the system dbus on the host.

### DIFF
--- a/src/dbus_fast/auth.py
+++ b/src/dbus_fast/auth.py
@@ -68,6 +68,8 @@ class AuthExternal(Authenticator):
     def _authentication_start(self, negotiate_unix_fd: bool = False) -> str:
         self.negotiate_unix_fd = negotiate_unix_fd
         uid = self.uid
+        if uid == -1:
+            return "AUTH EXTERNAL"
         if uid is None:
             uid = os.getuid()
         hex_uid = str(uid).encode().hex()
@@ -85,6 +87,9 @@ class AuthExternal(Authenticator):
 
         if response is _AuthResponse.AGREE_UNIX_FD:
             return "BEGIN"
+
+        if response is _AuthResponse.DATA and self.uid == -1:
+            return _AuthResponse.DATA
 
         raise AuthError(f"authentication failed: {response.value}: {args}")
 
--- a/src/dbus_fast/auth.py
+++ b/src/dbus_fast/auth.py
@@ -59,6 +59,9 @@ class AuthExternal(Authenticator):
     """An authenticator class for the external auth protocol for use with the
     :class:`MessageBus <dbus_fast.message_bus.BaseMessageBus>`.
 
+    :param uid: The uid to use when connecting to the message bus. Use UID_NOT_SPECIFIED to use the uid known to the kernel.
+    :vartype uid: int
+
     :sealso: https://dbus.freedesktop.org/doc/dbus-specification.html#auth-protocol
     """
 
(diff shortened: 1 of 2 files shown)


src/dbus_fast/auth.py
```diff
@@ -68,6 +68,8 @@ def __init__(self, uid: int = None) -> None:
     def _authentication_start(self, negotiate_unix_fd: bool = False) -> str:
         self.negotiate_unix_fd = negotiate_unix_fd
         uid = self.uid
+        if uid == -1:
```
bdraco, line 71: Does this have to be `-1`?

mvn23: Nope, but it seemed reasonable to take an int that would not be valid as a uid.

src/dbus_fast/auth.py
```diff
@@ -86,6 +88,9 @@ def _receive_line(self, line: str) -> str:
         if response is _AuthResponse.AGREE_UNIX_FD:
             return "BEGIN"
 
+        if response is _AuthResponse.DATA and self.uid == -1:
+            return _AuthResponse.DATA
```
mvn23, line 92: Not sure if we should return a bare string here instead.

src/dbus_fast/auth.py
bdraco: Lets make it a named constant so it can be changed later under the hood without having a breaking change

bdraco:
Can you add coverage for these lines?
Screenshot 2023-01-05 at 1 59 18 PM

mvn23:
All done. Should we try to fall back to this method if the default method with the uid from `os.getuid()` fails?

bdraco:
> All done. Should we try to fall back to this method if the default method with the uid from `os.getuid()` fails?

I think its probably fine without it `os.getuid()` is unlikely to fail

bdraco:
This looks good.

Would you update the doc string below as well?

```class AuthExternal(Authenticator):
    """An authenticator class for the external auth protocol for use with the
    :class:`MessageBus <dbus_fast.message_bus.BaseMessageBus>`.
    :sealso: https://dbus.freedesktop.org/doc/dbus-specification.html#auth-protocol
    """
```

mvn23:
> > All done. Should we try to fall back to this method if the default method with the uid from `os.getuid()` fails?
> 
> I think its probably fine without it `os.getuid()` is unlikely to fail

It is currently the only reason I'm patching the Home Assistant docker image locally...
Granted, it's quite a specific scenario (running containerized in a user namespace with mapped uids), but if we only fall back to it _after_ the initial attempt with the uid has failed, what's the harm?

bdraco:
> > I think its probably fine without it `os.getuid()` is unlikely to fail
> 
> It is currently the only reason I'm patching the Home Assistant docker image locally... Granted, it's quite a specific scenario (running containerized in a user namespace with mapped uids), but if we only fall back to it _after_ the initial attempt with the uid has failed, what's the harm?

Isn't that silently hiding an issue, it might not be the behavior the caller wants?

bdraco (approved):

bdraco:
Tested 👍

PR #197, opened 2023-02-20
fix: corrects Variant documentation

Mocramis:
Variant documentation does not match the class signature since the fork from dbus-python.

It troubled me while reading the doc, so here is the (very small) fix.


Mocramis:
Removed three lines of comments so coverage reduced :/ not sure how to fix that.

bdraco (approved):

bdraco:
Thanks @Mocramis

PR #199, opened 2023-04-20
feat: improve unmarshall performance

bdraco:
This works out to ~3.7% speed up against the unmarshall performance test


bdraco:
tested in production with bleak

PR #202, opened 2023-05-12
fix: subpath bad matching

Mocramis:
Although uncommon, i saw nothing in the dbus spec forbidding sub-path matching possible parent paths.

Current implementation of Introspect assumes this is the case though resulting in path being presented where they are not for instance instrospect on `/a` returning subnodes `b` and `d` when the registered paths are `/a/b` and `/c/a/d` 

This mr tests and fixes the issue.

Note: a more efficient way to fix would have been to use [str.removeprefix](https://docs.python.org/3/library/stdtypes.html?highlight=removeprefix#str.removeprefix) but it is only available from python3.9.

### DIFF
--- a/src/dbus_fast/message_bus.py
+++ b/src/dbus_fast/message_bus.py
@@ -673,11 +673,10 @@ class BaseMessageBus:
         children = set()
 
         for export_path in self._path_exports:
-            try:
-                child_path = export_path.split(path, maxsplit=1)[1]
-            except IndexError:
+            if not export_path.startswith(path):
                 continue
 
+            child_path = export_path.split(path, maxsplit=1)[1]
             child_path = child_path.lstrip("/")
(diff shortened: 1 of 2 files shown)


Mocramis:
Hmm i'm not sure what Codecov means by "Consider uploading the report"

bdraco:
I think this change is fine. I would like Mike to take a look at it though since it has potential impact to supervisor

Mocramis:
> I think this change is fine. I would like Mike to take a look at it though since it has potential impact to supervisor

Any news ?

bdraco:
I chatted with Mike about this earlier this week. He hasn't had time to test it yet

Mocramis:
any updates ?

bdraco (approved):

bdraco:
This LGTM. I can't see this causing a problem.

PR #204, opened 2023-05-26
feat: speed up handler lookup

bdraco:


src/dbus_fast/message_bus.py
```diff
@@ -905,45 +908,45 @@ def handler(msg: Message, send_reply: Callable[[Message], None]) -> None:
...
+                            and msg.signature == method.in_signature
+                        ):
+                            handler = _SERVICE_INTERFACE_GET_HANDLER(
+                                interface, method, self
+                            )
+                            if handler:
+                                return handler
 
-        return handler
+        return None
```
bdraco, line 949: Looks like everything is covered except returning `None` 👍

bdraco:
improvement is still in the margin of error

PR #205, opened 2023-07-02
feat: cache unmarshall when messages are small

bdraco:


bdraco:
neat idea, but probably not going to work due to serial

bdraco:
got it working but because of the copy in pure python its slower

PR #211, opened 2023-07-27
fix: avoid raising exception when we know there is no data in the buffer

bdraco:
If we know `_unmarshall` will internally raise `EOFError` via `peek`, we now return and wait until asyncio calls us again to read

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,12 @@
 
 <!--next-version-placeholder-->
 
+## v1.87.1 (2023-07-24)
+
+### Fix
+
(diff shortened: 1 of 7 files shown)


bdraco:
peek does more than peek

It can actually READ!

https://github.com/python/cpython/blob/f106254e0b3c83d9064763d6132b3fe997da901b/Lib/_pyio.py#L1151

bdraco:
this is so much better for small messages but kills performance for large ones

bdraco:
I'm not sure this is going to work since there is no public method to peek that does not cause a read to happen

bdraco:
Its not worth the change risk unless we can actually avoid the syscall

PR #213, opened 2023-07-27
fix: avoid double buffering when using asyncio reader without negotiate_unix_fd

bdraco:
We already buffer in the unmarshaller so we shouldn't do it twice if we can avoid it when `negotiate_unix_fd` is `False`

`recvmsg` for the `negotiate_unix_fd` path did not check for `EAGAIN` or `EWOULDBLOCK` which could result in unexplained failures

This is ~33.17% speed up

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -72,7 +72,7 @@ jobs:
             REQUIRE_CYTHON=1 poetry install --only=main,dev
           fi
       - name: Test with Pytest
-        run: export $(dbus-launch); poetry run pytest --cov-report=xml --timeout=5
+        run: export $(dbus-launch); poetry run pytest  -vvvs --cov-report=xml --timeout=5
       - name: Upload coverage to Codecov
         uses: codecov/codecov-action@v3
         with:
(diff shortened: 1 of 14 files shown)


bdraco:
This is going to be slower because recvmsg is called instead of `recv`

We could add a new path for `recv`

bdraco:
This is better but still not great for large messages (not that dbus usually sends large messages)

We should also fix recvmsg to try multiple times if it doesn't fill the wanted size

bdraco:
Screenshot 2023-07-27 at 4 28 26 PM


We git rid of this but the cython block is larger.

Its still not much better

bdraco:
We can do better. Lets see

bdraco:
Ran the profile 10 more times. Its definitely about 30% better

bdraco:
We can split this into 3 readers 

1. stream reader (still used for glib)
2. reader with fds (needs loop)
3. standard recv (needs loop)

bdraco:
With 3 different reader functions we can optimize for each use case.

We should be able to avoid the extra read since we know when we are at the end of the message. That could reduce the system calls by 33%

bdraco:
We need read to pos to over fill the buffer and reset to chop off the start of the old message so nothing ever stays in the stream.  With that, we still have optimized reads but when we have a full messages we don't read and wait for select if there is nothing in the buffer.  

That avoids the double buffer and copy and all the extra syscalls. Likely 30-40% speed up

PR #219, opened 2023-08-02
fix: more cython3 optional fixes

bdraco:
there were a few more places that were missing optional which is now strictly enforced with cython3

### DIFF
--- a/src/dbus_fast/introspection.py
+++ b/src/dbus_fast/introspection.py
@@ -172,8 +172,8 @@ class Method:
         assert_member_name_valid(name)
 
         self.name = name
-        self.in_args = in_args or []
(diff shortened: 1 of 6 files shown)


src/dbus_fast/introspection.py
```diff
@@ -172,8 +172,8 @@ def __init__(self, name: str, in_args: List[Arg] = [], out_args: List[Arg] = [])
         assert_member_name_valid(name)
 
         self.name = name
-        self.in_args = in_args
-        self.out_args = out_args
+        self.in_args = in_args or []
+        self.out_args = out_args or []
```
bdraco, line 176:
```suggestion
        self.in_args = in_args
        self.out_args = out_args
```

PR #220, opened 2023-08-05
fix: spelling of `dbus_fast.auth.AuthAnnonymous` to `dbus_fast.auth.AuthAnonymous`

jimduchek:
Pretty straightforward.  Don't think very many people use the ANONYMOUS auth, but included a line to keep backwards compatibility to not break anybody using the misspelled version.


bdraco:
This looks fine. Please add a test to make sure both names work so we do not regress this in the future by accident

jimduchek:
Good call.  Will add that now.

jimduchek:
Think that's as good as it's gonna get... the codecov check is failing in a file I didn't touch, so not sure what that's about.

bdraco (approved):

bdraco:
thanks @jimduchek

PR #221, opened 2023-08-09
feat: speed up to processing bluez passive data

bdraco:
This is a ~11% speed up

### DIFF
--- a/src/dbus_fast/message_bus.py
+++ b/src/dbus_fast/message_bus.py
@@ -43,10 +43,10 @@ def _block_unexpected_reply(reply: _Message) -> None:
     """Block a reply if it's not expected.
 
     Previously we silently ignored replies that were not expected, but this
-    lead to implementation errors that were hard to debug. Now we log an
-    error instead.
+    lead to implementation errors that were hard to debug. Now we log a
+    debug message instead.
     """
-    _LOGGER.error(
+    _LOGGER.debug(
         "Blocked attempt to send a reply from handler "
         "that received a message with flag "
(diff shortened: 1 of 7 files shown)


src/dbus_fast/message_bus.py
```diff
@@ -892,8 +922,11 @@ def _make_method_handler(
         def _callback_method_handler(
             msg: Message, send_reply: Callable[[Message], None]
         ) -> None:
+            result = method_fn(interface, *msg_body_to_args(msg))
+            if not _expects_reply(msg):
+                return
```
bdraco, line 927: The default was to always call send_reply when a reply wasn't expected it would just be thrown away. All that work for 6500 DeviceFound messages per minute and they are just discarded 🙈

src/dbus_fast/message_bus.py
```diff
@@ -27,13 +27,34 @@
...
+_LOGGER = logging.getLogger(__name__)
+
 
 _Message = Message
 
 
 def _expects_reply(msg: _Message) -> bool:
-    return not (msg.flags.value & NO_REPLY_EXPECTED_VALUE)
+    """Whether a message expects a reply."""
+    return not (msg.flags & NO_REPLY_EXPECTED)
```
bdraco, line 39: This turned out to be faster with pure python by slower with cython. I'm going to revert it

bdraco: https://github.com/Bluetooth-Devices/dbus-fast/releases/tag/v1.91.1

PR #225, opened 2023-08-17
fix: messages could be sent out of order if they had to queue

bdraco:
The queue should have been read from the left, but it was instead read from the right which meant if
the message did not send right away it was LIFO instead of FIFO

This went unnoticed because most of the time we can send immediately and it never queues

### DIFF
--- a/src/dbus_fast/aio/message_bus.py
+++ b/src/dbus_fast/aio/message_bus.py
@@ -48,7 +48,7 @@ class _MessageWriter:
         self.bus = bus
         self.sock = bus._sock
         self.loop = bus._loop
-        self.buf: Optional[bytearray] = None
+        self.buf: Optional[memoryview] = None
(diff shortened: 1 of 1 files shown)


src/dbus_fast/aio/message_bus.py
```diff
@@ -37,28 +37,38 @@ def _future_set_result(fut: asyncio.Future, result: Any) -> None:
...
+                    # If there is no buffer, get the next message
                     if not self.messages:
                         # nothing more to write
                         if remove_writer:
                             self.loop.remove_writer(self.fd)
                         return
-                    buf, unix_fds, fut = self.messages.pop()
+
+                    # Get the next message
+                    buf, unix_fds, fut = self.messages.popleft()
```
bdraco, line 71: We needed to popleft

PR #232, opened 2023-08-24
fix: avoid cythonizing SendReply

### LINKED ISSUE #229: Cython's issue with python3.9 since dbus-fast v1.90.2
**The bug**
When a DBusError is raised from a method, the raise exception `_as_message`  seems to not be properly resolved from dbus_fast.message_bus.SendReply._exit. It leads to no answer ever being sent on the dbus-message.

The error does not happen without cython.

**To Reproduce**
With the following test file (which is a trimmed down version of the high-level service example from the doc):
```python
#test_bug.py
from dbus_fast import DBusError
from dbus_fast.aio import MessageBus
from dbus_fast.service import ServiceInterface, method


    def __init__(self):
        super().__init__('com.example.SampleInterface0')

    @method()
    async def fail(self):
        raise DBusError('com.example.name.fail', 'fail')


async def start_service():
    bus0 = await MessageBus().connect()
    iface = ExampleInterface()
    bus0.export('/com/example/sample0', iface)
    await bus0.request_name('com.example.name')


async def call_fail():
    bus = await MessageBus().connect()
    introspection = await bus.introspect('com.example.name',
                                         '/com/example/sample0')

    proxy_object = bus.get_proxy_object('com.example.name',
                                        '/com/example/sample0',
                                        introspection)
    interface = proxy_object.get_interface('com.example.SampleInterface0')
    await interface.call_fail()


async def startncall():
    await start_service()
    await call_fail()


if __name__ == "__main__":
    import asyncio
    asyncio.run(startncall())
else:
    import pytest

    @pytest.mark.asyncio
    async def test_fail_bug():
        with pytest.raises(DBusError):
            await startncall()
```

In a venv with dbus-fast >=1.90.2 run:
```bash
$ python3.9 test_bug.py
Exception in callback MessageBus._make_method_handler.<locals>._coroutine_method_handler.<locals>._done(<Task finishe...Error('fail')>) at /.cache/pypoetry/virtualenvs/dbus-fast-b0n82em1-py3.9/lib/python3.9/site-packages/dbus_fast/aio/message_bus.py:456
handle: <Handle MessageBus._make_method_handler.<locals>._coroutine_method_handler.<locals>._done(<Task finishe...Error('fail')>) at /.cache/pypoetry/virtualenvs/dbus-fast-b0n82em1-py3.9/lib/python3.9/site-packages/dbus_fast/aio/message_bus.py:456>
  File "/.cache/pypoetry/virtualenvs/dbus-fast-b0n82em1-py3.9/lib/python3.9/site-packages/dbus_fast/aio/message_bus.py", line 459, in _done
    result = fut.result()
  File "/misc/dbus-fast/test_bug.py", line 12, in fail
    raise DBusError('com.example.name.fail', 'fail')
dbus_fast.errors.DBusError: fail

During handling of the above exception, another exception occurred:

  File "/usr/lib/python3.9/asyncio/events.py", line 80, in _run
    self._context.run(self._callback, *self._args)
  File "/.cache/pypoetry/virtualenvs/dbus-fast-b0n82em1-py3.9/lib/python3.9/site-packages/dbus_fast/aio/message_bus.py", line 463, in _done
    send_reply(
  File "src/dbus_fast/message_bus.py", line 103, in dbus_fast.message_bus.SendReply.__exit__
    return self._exit(exc_type, exc_value, tb)
  File "src/dbus_fast/message_bus.py", line 84, in dbus_fast.message_bus.SendReply._exit
    self(exc_value._as_message(self._msg))
AttributeError: type object 'Exception' has no attribute '_as_message'
```

The same problem happens with `pytest test_bug.py`, however i cannot trigger the issue when running with `poetry run pytest` as if the corresponding dbus-fast was without cython compilation.

**Additional context**

Only reproduced with cython and with python3.9 (not python3.11). 

It is not caught by the tests suite because it seems the task_method testsuite only run in non-cython mode. (Or something else is happening, i do not know really well how poetry works).

The bug happens since v1.90.2 but is not linked to this update only to the pipy binary: i guess the cython version used to compile changed at this point (I did not manage to track down the culprit cython version)

bdraco:
This is a workaround since context managers currently have an issue with Cython 3

fixes #229

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -96,5 +96,5 @@ module = "docs.*"
 ignore_errors = true
 
 [build-system]
-requires = ['setuptools>=65.4.1', 'wheel', 'Cython<3.0.0', "poetry-core>=1.0.0"]
+requires = ['setuptools>=65.4.1', 'wheel', 'Cython', "poetry-core>=1.0.0"]
 build-backend = "poetry.core.masonry.api"
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -96,5 +96,5 @@ module = "docs.*"
 ignore_errors = true
 
 [build-system]
-requires = ['setuptools>=65.4.1', 'wheel', 'Cython', "poetry-core>=1.0.0"]
+requires = ['setuptools>=65.4.1', 'wheel', 'Cython>=3.0.0', "poetry-core>=1.0.0"]
 build-backend = "poetry.core.masonry.api"
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -96,5 +96,5 @@ module = "docs.*"
 ignore_errors = true
 
 [build-system]
-requires = ['setuptools>=65.4.1', 'wheel', 'Cython>=3.0.0', "poetry-core>=1.0.0"]
+requires = ['setuptools>=65.4.1', 'wheel', 'Cython<3.0.0', "poetry-core>=1.0.0"]
 build-backend = "poetry.core.masonry.api"
--- a/src/dbus_fast/message_bus.pxd
+++ b/src/dbus_fast/message_bus.pxd
@@ -18,11 +18,6 @@ cdef object assert_bus_name_valid
 cdef _expects_reply(Message msg)
 
 
-cdef class SendReply:
-
-    cdef object _bus
-    cdef object _msg
-
 cdef class BaseMessageBus:
 
     cdef public object unique_name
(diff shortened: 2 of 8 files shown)


bdraco:
Performance impact is effectively zero for Bluez DeviceFound messages since there is no reply 

This seems like it will be fine.

Tested on production and no side effects observed (wouldn't expect it from moving code around)

PR #236, opened 2023-09-01
fix: handling of None messages from notify callback

Mocramis:
This is especially common as BaseMessageBus._finalize calls handler(None, err) when an exception is raised.

From the surrounding code, it seems the common assumption is that msg should be None if err is set. However you might want to print msg if it is here even if there is an err, but then the logs would be awkward no ?

Did not take the time to build a test to reproduce the issue though

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -69,7 +69,7 @@ jobs:
             REQUIRE_CYTHON=1 poetry install --only=main,dev
           fi
       - name: Test with Pytest
-        run: export $(dbus-launch); poetry run pytest -vvvs --cov-report=xml --timeout=5 tests/client/test_methods.py::test_aio_proxy_object
+        run: export $(dbus-launch); poetry run pytest --cov-report=xml --timeout=5
       - name: Upload coverage to Codecov
         uses: codecov/codecov-action@v3
         with:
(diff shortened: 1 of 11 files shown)


bdraco:
Looks ok, but we do need a test before merging

Mocramis:
Hello, so i have a reproducer but i cannot automate it because this is triggered through a race condition and the side-effect is a log:
```python
import pytest

from dbus_fast import DBusError, aio, glib
from dbus_fast.service import ServiceInterface


    def __init__(self):
        super().__init__("test.interface")


@pytest.mark.asyncio
async def test_fast_disconnect():
    bus_name = "aio.client.test.methods"

    bus = await aio.MessageBus().connect()
    bus2 = await aio.MessageBus().connect()
    await bus.request_name(bus_name)
    service_interface = ExampleInterface()
    bus.export("/test/path", service_interface)
    introspection = await bus2.introspect(bus_name, "/test/path")
    bus2.get_proxy_object(bus_name, "/test/path", introspection)
    bus2.disconnect()
    raise Exception("Failed")
```
This reproduce the error if you look at the captured stderr but if i try to use caplog to capture the logs, the error suddenly disapear.

Any thoughs about this ?

thanks.

bdraco:
Even if the test doesn't fail before it's still useful in case we break something else in the future as this project is definitely lighter on coverage than I would like

Mocramis:
The timeout is not the issue, i fail to understand how i broke 3.8 and 3.9, i do not reproduce the issue locally with my 3.9 python version :/

bdraco (approved):
Thanks @Mocramis

PR #238, opened 2023-09-06
feat: small speed up to the unmarshaller

bdraco:


bdraco:
tested on production 👍

Issue #242, opened 2023-09-07
Allow use without importing `glib` backend.


dlech:
**Is your feature request related to a problem? Please describe.**
This came up in https://github.com/hbldh/bleak/issues/1412

In addition to avoiding a crash like in this case, it would be nice to improve startup time and reduce memory usage on constrained systems where `gi` is installed.

**Describe the solution you'd like**
The top-level `__init__.py` should not import `aio` or `glib`. It should be up to the library user to import the backend they want to use.

**Additional context**
I also wouldn't be opposed to dropping the `glib` backend completely and making this library asyncio-only.

bdraco:
Thats a good idea.  I'm not sure if we should drop glib completely though, but definitely run time load

PR #243, opened 2023-09-07
feat!: don't import backends by default

dlech:
Since the top-level `__init__.py` is always imported when any subpackage is imported, this caused both backends (glib and aio) to always be imported. This is not desirable, since it means that the glib backend will always be imported, even if the user only wants to use the aio. This wastes resources and in some cases can even cause a crash.

Fixes: https://github.com/Bluetooth-Devices/dbus-fast/issues/242

### DIFF
--- a/src/dbus_fast/__init__.py
+++ b/src/dbus_fast/__init__.py
@@ -1,4 +1,4 @@
-from . import aio, glib, introspection, message_bus, proxy_object, service
+from . import introspection, message_bus, proxy_object, service
 from .constants import (
     ArgDirection,
     BusType,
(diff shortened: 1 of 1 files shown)


dlech:
Technically, this is an API breaking change, but I think the only thing it would break is:

```python
import dbus_fast

dbus_fast.aio.something
```

bdraco:
We should probably mark it as a breaking change.

Testing this now

bdraco (approved):

bdraco:
Pushed to all my production. Nothing broke. Let's hope thats the case for everyone else

bdraco:
Thanks @dlech

bdraco:
Everything was ok with the dep bump except `bleak`

`    bleak 0.21.0 depends on dbus-fast<2.0.0 and >=1.83.0; platform_system == "Linux"`

PR #253, opened 2023-09-13
feat: speed up readers in the unmarshall path

bdraco:


bdraco:
CI seems happy. cython had a lot of bugs with `cdef void` before but they all appear to be fixed 🤞

PR #254, opened 2023-09-20
feat: speed up unmarshalling Variants

bdraco:
This is a 3.5% speed up, but we can apply the same idea in a few more places


bdraco:
production testing looks good

PR #266, opened 2023-11-07
fix: re-release since the previous release ran out of space on PyPI

bdraco:
I had to delete some old releases

https://pypi.org/help/#project-size-limit


bdraco:
I think I deleted enough old releases to get by until https://github.com/pypi/support/issues/3341

hopefully nobody is still using the pinned versions but I figured thats better than having a release out there which doesn't have wheels

PR #267, opened 2023-11-10
feat: add support for tuples to the marshaller

bdraco:
fixes test failures seen in https://github.com/home-assistant/supervisor/pull/4681


bdraco:
🤞 that I deleted enough old released (and picked ones that nobody has pinned to) that this publishes ok since https://github.com/pypi/support/issues/3341 is pending

PR #268, opened 2023-11-21
feat: make string enum constants StrEnums

bdraco:


bdraco:
Apparently an arbitrary string is allowed here so this won't work

PR #276, opened 2023-12-12
feat: speed up message callbacks

bdraco:


src/dbus_fast/message_bus.py
```diff
@@ -901,7 +908,7 @@ def _make_method_handler(
     def _find_message_handler(
         self, msg: _Message
     ) -> Optional[Callable[[Message, Callable[[Message], None]], None]]:
-        if msg.interface.startswith("org.freedesktop.DBus."):
```
bdraco, line 904: There is no optimized version of this in cython so it has to call the actual python function

src/dbus_fast/message_bus.py
```diff
@@ -901,7 +908,7 @@ def _make_method_handler(
...
+        if "org.freedesktop.DBus." in msg.interface:
```
bdraco, line 911: This can use the Pyx version

src/dbus_fast/message_bus.py
```diff
@@ -37,6 +39,11 @@
 
 def _expects_reply(msg: _Message) -> bool:
     """Whether a message expects a reply."""
+    if msg.flags is NO_REPLY_EXPECTED:
+        return False
+    if msg.flags is NONE:
+        return True
```
bdraco, line 45: 99% of cases can avoid the python jump here

PR #280, opened 2024-05-04
fix: introspection bogus child paths

### LINKED ISSUE #279: Introspection returns bogus child paths
**Describe the bug**

I see the following path hierarchy when showing the object tree with busctl tree `com.example.Service`
```
└─/com
  └─/com/example
    └─/com/example/service
      ├─/com/example/service/item
      │ ├─/com/example/service/item/0
      │ ├─/com/example/service/item/1
      │ │ ├─/com/example/service/item/1/0
      │ │ └─/com/example/service/item/1/1
      │ ├─/com/example/service/item/10
      │ ├─/com/example/service/item/11
      │ └─/com/example/service/item/2
...
```

I would expect to see only
```
└─/com
  └─/com/example
    └─/com/example/service
      ├─/com/example/service/item
      │ ├─/com/example/service/item/0
      │ ├─/com/example/service/item/1
      │ ├─/com/example/service/item/10
      │ ├─/com/example/service/item/11
      │ └─/com/example/service/item/2
...
```

The `/com/example/service/item/1/0` and `/com/example/service/item/1/1` are bogus and have never been exported by the service. They also show up without any interfaces when being introspected with busctl introspect.

It looks like that the computation of the children nodes in https://github.com/Bluetooth-Devices/dbus-fast/blob/f0fc6668b9543bd81e2438fe78a54c31702a4bc5/src/dbus_fast/message_bus.py#L650 is wrong if some paths are prefixes for other paths.  For example, `/com/example/service/item/1` is a prefix of `/com/example/service/item/10`, but the latter is not a child of the former.

**To Reproduce**

* Expose object paths where one path is a prefix of another path but they have no parent-child relationship. 
* Introspect with busctl tree

**Additional context**

Suggested fix: Replace https://github.com/Bluetooth-Devices/dbus-fast/blob/f0fc6668b9543bd81e2438fe78a54c31702a4bc5/src/dbus_fast/message_bus.py#L663 with

```py
            if not export_path.startswith(path + '/'):
```

mriemensberger:
fixes #279

### DIFF
--- a/tests/service/test_standard_interfaces.py
+++ b/tests/service/test_standard_interfaces.py
@@ -91,6 +91,7 @@ async def test_introspect_matching_sub_paths():
     interface = ExampleInterface("test.interface1")
 
     bus1.export("/a/test/path1", interface)
+    bus1.export("/a/test/path10", interface)
     bus1.export("/a/subpath/a/test/path2", interface)
 
     async def introspect_subpath(path, expected_subnodes):
@@ -108,8 +109,9 @@ async def test_introspect_matching_sub_paths():
 
     await introspect_subpath("/", {"a"})
     await introspect_subpath("/a", {"test", "subpath"})
-    await introspect_subpath("/a/test", {"path1"})
+    await introspect_subpath("/a/test", {"path1", "path10"})
     await introspect_subpath("/a/test/path1", set())
+    await introspect_subpath("/a/test/path10", set())
     await introspect_subpath("/a/subpath/a/test", {"path2"})
     await introspect_subpath("/a/subpath/a/test/path2", set())
 
--- a/src/dbus_fast/message_bus.py
+++ b/src/dbus_fast/message_bus.py
@@ -664,6 +664,9 @@ class BaseMessageBus:
                 continue
 
             child_path = export_path.split(path, maxsplit=1)[1]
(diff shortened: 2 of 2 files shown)


tests/service/test_standard_interfaces.py
```diff
@@ -108,8 +109,9 @@ async def introspect_subpath(path, expected_subnodes):
 
     await introspect_subpath("/", {"a"})
     await introspect_subpath("/a", {"test", "subpath"})
-    await introspect_subpath("/a/test", {"path1"})
+    await introspect_subpath("/a/test", {"path1", "path10"})
     await introspect_subpath("/a/test/path1", set())
```
bdraco, line 113: It looks like we need some more coverage for passing paths with a trailing slash since the new check above it doing `path and path[-1] != "/"`

mriemensberger: I think `/`is the only valid object path with a trailing slash. I adapted the check to only test `path != "/"`.

bdraco (approved):

bdraco:
Thanks @mriemensberger 

Changes LGTM.  I'll leave this for other reviewers for a few days and than do a release with the fix if there are no comments

PR #281, opened 2024-05-20
fix: clear exception flag on disconnect future if its also sent to handlers

bdraco:
If we do not clear the exception retrieved flag on the disconnect future it will likely never be retrieved since wait_for_disconnect will never be called when the reply handler raises.

### DIFF
--- a/src/dbus_fast/aio/message_bus.py
+++ b/src/dbus_fast/aio/message_bus.py
@@ -1,5 +1,6 @@
 import array
 import asyncio
+import contextlib
 import logging
 import socket
 from collections import deque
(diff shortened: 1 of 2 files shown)


src/dbus_fast/aio/message_bus.py
```diff
@@ -522,13 +523,23 @@ def _finalize(self, err: Optional[Exception] = None) -> None:
...
+        had_handlers = bool(self._method_return_handlers or self._user_message_handlers)
+
         super()._finalize(err)
 
         if self._disconnect_future.done():
             return
 
         if err and not self._user_disconnect:
             _future_set_exception(self._disconnect_future, err)
+            # If this happens during a reply the message handlers
```
bdraco, line 535:
```suggestion
            # If this happens during a reply, the message handlers
```

PR #283, opened 2024-06-26
feat: build wheels for aarch64 to allow use in embedded systems

Mimoja:
As github runners can be used with qemu we simply emulate the aarch64 cpu needed to build for it. This allows faster installations on tradionally weaker embedded hardware.
The full aarch64 run roughly takes  90 minutes which should be taken into consideration if it is worth the occasional but additional time investment on release.


Mimoja:
I dont think the python 3.7 failures are due to the PR. They also pop up on the previous head
https://github.com/Mimoja/dbus-fast/actions/runs/9673188373/job/26686694693

bdraco:
I'll rip out 3.7 support later today if my flight isn't delayed

bdraco:
> As github runners can be used with qemu we simply emulate the aarch64 cpu needed to build for it. This allows faster installations on tradionally weaker embedded hardware.
> 
> The full aarch64 run roughly takes  90 minutes which should be taken into consideration if it is worth the occasional but additional time investment on release.

It's no problem to do the aarch64 as well since the shell builds happen after the release and get uploaded lazily

bdraco (approved):

bdraco:
first build failed. I think https://github.com/Bluetooth-Devices/dbus-fast/pull/285 will fix it

bdraco:
new attempt is https://github.com/Bluetooth-Devices/dbus-fast/actions/runs/9674505623

Mimoja:
Awesome. Thank you!

Mimoja:
Can confirm install working now perfectly fine on aarch64 musl and glibc. And it can now be installed on low memory systems which previously failed \0/

PR #291, opened 2024-08-21
feat: python 3.13 support

bdraco:


.github/workflows/ci.yml
```diff
@@ -15,7 +15,7 @@ jobs:
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v3
-      - uses: actions/setup-python@v3
+      - uses: actions/setup-python@v4
```
dlech, line 18: why not v5?

bdraco, line 18:
```suggestion
      - uses: actions/setup-python@v5
```

bdraco: copied from another template and didn't update.  updated now

bdraco:
Probably ok to merge but I need to babysit the wheel builds and its too late to do that since I need to get up early for a flight. Will have to do it later this week when I have more time.

PR #299, opened 2024-08-22
feat: use dbus-run-session to drop X11 dependency

agners:
The dbus-run-session helper allows to run a session scoped D-Bus daemon without X11 dependency.


bdraco (approved):

bdraco:
will merge shortly after I cleanup the dep bumps

bdraco:
Thanks @agners

PR #264, opened 2023-11-07
feat: improve marshaller performance

bdraco:
This is roughly a 10% speed up marshalling vs the benchmark

adds some more missing types

PR #310, opened 2024-09-06
fix: add missing cython version pin to the build system

bdraco:
related comment: https://github.com/Bluetooth-Devices/dbus-fast/pull/264#issuecomment-2334550288

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -101,5 +101,5 @@ module = "docs.*"
 ignore_errors = true
 
 [build-system]
-requires = ['setuptools>=65.4.1', 'wheel', 'Cython', "poetry-core>=1.0.0"]
+requires = ['setuptools>=65.4.1', 'wheel', 'Cython>=0.29.32,<3.1.0', "poetry-core>=1.0.0"]
 build-backend = "poetry.core.masonry.api"


pyproject.toml
```diff
@@ -101,5 +101,5 @@ module = "docs.*"
 ignore_errors = true
 
 [build-system]
-requires = ['setuptools>=65.4.1', 'wheel', 'Cython', "poetry-core>=1.0.0"]
+requires = ['setuptools>=65.4.1', 'wheel', 'Cython>=0.29.32,<3.1.0', "poetry-core>=1.0.0"]
```
jameshilliard, line 104: The [error](https://github.com/Bluetooth-Devices/dbus-fast/pull/264#issuecomment-2334550288) was hit with Cython version `0.29.37`, so I don't think this is accurate.

bdraco: Yeah it looks like we need 3+

bdraco: https://github.com/Bluetooth-Devices/dbus-fast/pull/311

bdraco:
https://github.com/Bluetooth-Devices/dbus-fast/releases/tag/v2.24.2 should appear in a hour or so

PR #329, opened 2024-11-14
validators: allow spaces in dbus property names

FakeShell:
pretty self explanatory. dbus property names can have spaces in them (its a legal character), an example is mmsd and all its implementations (mmsd-tng and mmsd4ofono)
https://gitlab.com/kop316/mmsd/-/blob/master/doc/message-api.txt?ref_type=heads#L66

a similar patch was applied to dbus-next for my mmsd4ofono implementation
property:
https://github.com/FuriLabs/mmsd4ofono/blob/trixie/mmsd/ofono_mms_message.py#L97

dbus next patch:
https://github.com/FuriLabs/python-dbus-next/commit/558fdec0b67cca2e1455b9c4801e21488afaffed


src/dbus_fast/validators.py
```diff
@@ -10,7 +10,7 @@
 
 _bus_name_re = re.compile(r"^[A-Za-z_-][A-Za-z0-9_-]*$")
 _path_re = re.compile(r"^[A-Za-z0-9_]+$")
-_element_re = re.compile(r"^[A-Za-z_][A-Za-z0-9_]*$")
+_element_re = re.compile(r"^[A-Za-z_][A-Za-z0-9_ ]*$")
```
bdraco, line 13: Can you point out where the spec allows spaces?

It doesn't appear to based on how I'm reading it

https://dbus.freedesktop.org/doc/dbus-specification.html
> Each element must only contain the ASCII characters "[A-Z][a-z][0-9]_"

bdraco: Maybe its OK to accept this change (assuming we also get some tests)

I asked other members to take a look in case there is a security impact I'm not thinking about

dlech (commented):
If the D-Bus properties in question work with other D-Bus implementations then I would say, we should allow it too. Also, property names are just strings passed as args to methods (or as keys in a dict), so I don't think they fall under the spec requirements for interface names or method names. So instead of changing `_element_re`, maybe we should just not use that to validate property names?

bdraco:
> So instead of changing `_element_re`, maybe we should just not use that to validate property names?

`_element_re` is currently only used for `is_interface_name_valid`

Maybe is would be sufficient to only change the validate for the last element?

black-desk:
Related: #358

PR #357, opened 2025-01-07
fix: revert avoid building wheels if a release is not made

bdraco:
Reverts Bluetooth-Devices/dbus-fast#355


bdraco:
We need to update to a newer PSR before this will work correctly

PR #358, opened 2025-01-10
fix: void validate arguments/properties name

black-desk:
According to the D-Bus specification, the parameters of methods and the
properties of interfaces are not “members” and should not be validated.

> ...
>
> #### Member names
>
> Member (i.e. method or signal) names:
>
> - Must only contain the ASCII characters "[A-Z][a-z][0-9]_" and may
>   not begin with a digit.
>
> - Must not contain the '.' (period) character.
>
> - Must not exceed the maximum name length.
>
> - Must be at least 1 byte in length.
>
> ...

https://dbus.freedesktop.org/doc/dbus-specification.html#message-protocol-names-member

Some realistic examples:

``` xml
  <interface name=“org.freedesktop.portal.PowerProfileMonitor”>
    <property type=“b” name=“power-saver-enabled” access=“read” />
    <property type=“u” name=“version” access=“read” />
  </interface
```

``` xml
  <interface name="org.gtk.Application">
    <method name="Activate">
      <arg type="a{sv}" name="platform-data" direction="in">
      </arg>
    </method>
    <method name="Open">
      <arg type="as" name="uris" direction="in">
      </arg>
      <arg type="s" name="hint" direction="in">
      </arg>
      <arg type="a{sv}" name="platform-data" direction="in">
      </arg>
    </method>
    <method name="CommandLine">
      <arg type="o" name="path" direction="in">
      </arg>
      <arg type="aay" name="arguments" direction="in">
      </arg>
      <arg type="a{sv}" name="platform-data" direction="in">
      </arg>
      <arg type="i" name="exit-status" direction="out">
      </arg>
    </method>
    <property type="b" name="Busy" access="read">
    </property>
  </interface>
```

### DIFF
--- a/src/dbus_fast/introspection.py
+++ b/src/dbus_fast/introspection.py
@@ -34,9 +34,6 @@ class Arg:
         direction: Optional[list[ArgDirection]] = None,
         name: Optional[str] = None,
     ):
-        if name is not None:
-            assert_member_name_valid(name)
-
         type_ = None
         if type(signature) is SignatureType:
             type_ = signature
@@ -246,8 +243,6 @@ class Property:
         signature: str,
         access: PropertyAccess = PropertyAccess.READWRITE,
     ):
-        assert_member_name_valid(name)
-
         tree = get_signature_tree(signature)
         if len(tree.types) != 1:
             raise InvalidIntrospectionError(


bdraco:
> Strictly speaking, D-Bus property names are not required to follow the same naming restrictions as member names, but D-Bus property names that would not be valid member names (in particular, GObject-style dash-separated property names) can cause interoperability problems and should be avoided.

bdraco:
I think the best path forward would be to split this PR into two PRs

1. disable validation of arguments. That seems fine as is

Since the spec warns about not conforming to member naming restrictions for properties, I think we should keep the validation on by default.

> Strictly speaking, D-Bus property names are not required to follow the same naming restrictions as member names, but D-Bus property names that would not be valid member names (in particular, GObject-style dash-separated property names) can cause interoperability problems and should be avoided.

2. add a `validate` param to `Property` that is enabled by default so validation can be turned off.  We already do this for [`Message`](https://github.com/Bluetooth-Devices/dbus-fast/blob/c3779d9247a73e77c6d5aa5260dffe7991f6fb15/src/dbus_fast/message.py#L124) so it would be consistent with that design.

cc @FakeShell

bdraco:
@black-desk Thanks for your contribution. Please see my comment above

black-desk:
> I think the best path forward would be to split this PR into two PRs
> ## 1 for the arguments
> 
> disable validation of arguments. That seems fine as is
> ## 2 for the properties
> 
> Since the spec warns about not conforming to member naming restrictions for properties, I think we should keep the validation on by default.
> 
> > Strictly speaking, D-Bus property names are not required to follow the same naming restrictions as member names, but D-Bus property names that would not be valid member names (in particular, GObject-style dash-separated property names) can cause interoperability problems and should be avoided.
> 
> Add a `validate` param to `Property` that is enabled by default so validation can be turned off. We already do this for [`Message`](https://github.com/Bluetooth-Devices/dbus-fast/blob/c3779d9247a73e77c6d5aa5260dffe7991f6fb15/src/dbus_fast/message.py#L124) so it would be consistent with that design.
> 
> cc @FakeShell

done.

But add validate to Property.__init__ is not enough, check my code for details.

bdraco:
> Tests will be updated later.

LGTM once tests are added

black-desk:
Some updates:

add validate_property_names to more introspect related functions:

- dbus_fast.aio.message_bus.MessageBus.introspect
- dbus_fast.introspection.Interface.from_xml
- dbus_fast.introspection.Node.parse
- dbus_fast.message_bus.BaseMessageBus.introspect

black-desk:
I note an issue while adding tests:

https://dbus.freedesktop.org/doc/dbus-specification.html#message-protocol-names-member says:

> Must only contain the ASCII characters "[A-Z][a-z][0-9]_" and may not begin with a digit.

but

https://github.com/Bluetooth-Devices/dbus-fast/blob/e7750caed5791aef0cbb8c62e82ccabf02f65df7/src/dbus_fast/validators.py#L14

bdraco:
> I note an issue while adding tests:
> 
> 
> 
> https://dbus.freedesktop.org/doc/dbus-specification.html#message-protocol-names-member says:
> 
> 
> 
> > Must only contain the ASCII characters "[A-Z][a-z][0-9]_" and may not begin with a digit.
> 
> 
> 
> but
> 
> 
> 
> https://github.com/Bluetooth-Devices/dbus-fast/blob/e7750caed5791aef0cbb8c62e82ccabf02f65df7/src/dbus_fast/validators.py#L14
> 
> 
> 
> 
> 
> 

That's unfortunate. Probably best to add a comment about it for now. It's likely too risky to change at this point.

black-desk:
> > Tests will be updated later.
> 
> LGTM once tests are added

Tests have been added. But I add validate_property_names to more methods, please review the change.

bdraco (approved):
Thanks @black-desk

src/dbus_fast/introspection.py
```diff
@@ -245,8 +242,10 @@ def __init__(
         name: str,
         signature: str,
         access: PropertyAccess = PropertyAccess.READWRITE,
+        validate: bool = True,
     ):
-        assert_member_name_valid(name)
+        if validate:
```
bdraco, line 247: Realized it after merging, but we should probably update the docstrings here as well

bdraco:
@mdegat01 also suggested adding some kind of doc note that says if you do have to turn off validation you probably should not use the introspection with the higher-level proxy objects

PR #360, opened 2025-01-10
feat: improve unmarshaller performance

bdraco:
Avoid some more `dict` lookups


bdraco:
~1%

not worth the change

Issue #315, opened 2024-09-17
Interface header should be optional


izak:
**Describe the bug**
According to the [dbus specification][1], the interface header can be omitted in method calls. If you do that, then this trace results:

```
ERROR    Unexpected error processing message: %s
Traceback (most recent call last):
File "/usr/lib/python3.8/site-packages/dbus_fast/aio/message_reader.py", line 23, in _message_reader
    process(message)
  File "/usr/lib/python3.8/site-packages/dbus_fast/message_bus.py", line 847, in _process_message
    handler = self._find_message_handler(msg)
  File "/usr/lib/python3.8/site-packages/dbus_fast/message_bus.py", line 918, in _find_message_handler
    if "org.freedesktop.DBus." in msg.interface:
TypeError: argument of type 'NoneType' is not iterable
```

This is the same bug I logged against [dbus_next][2].

**To Reproduce**
Run example service
```
python3 examples/example-service.py
```

Call method with no interface. For this I used the old python-dbus package, since this is also where my bug is coming from: Old code using that package.
```
import dbus
bus = dbus.SessionBus()
bus.call_blocking('dbus.next.example.service', '/example/path', None, 'Echo', '', ['hello'])
```

**Possible fix**
This fixes it for me, but I didn't have time to really absorb your whole test/pre-commit framework. I also don't know if this affects performance disproportionately.

```
diff --git a/src/dbus_fast/message_bus.py b/src/dbus_fast/message_bus.py
index 14776fb..aa9a7d5 100644
--- a/src/dbus_fast/message_bus.py
+++ b/src/dbus_fast/message_bus.py
@@ -915,7 +915,7 @@ class BaseMessageBus:
     def _find_message_handler(
         self, msg: _Message
     ) -> Optional[Callable[[Message, Callable[[Message], None]], None]]:
-        if "org.freedesktop.DBus." in msg.interface:
+        if msg.interface is not None and "org.freedesktop.DBus." in msg.interface:
             if (
                 msg.interface == "org.freedesktop.DBus.Introspectable"
                 and msg.member == "Introspect"
@@ -949,7 +949,10 @@ class BaseMessageBus:
                     if method.disabled:
                         continue

-                    if (
+                    if msg.interface is None and (
+                        msg.member == method.name
+                        and msg.signature == method.in_signature) \
+                    or (
                         msg.interface == interface.name
                         and msg.member == method.name
                         and msg.signature == method.in_signature

```

[...]

PR #390, opened 2025-02-12
feat: make interface optional on method calls

izak:
The dbus spec allows the interface to be optional on method calls. In that case, allow the message to match on any method with a matching name and signature.

https://github.com/Bluetooth-Devices/dbus-fast/issues/315


bdraco:
Would you please add a test case for this change

PR #400, opened 2025-03-05
feat: refactor service bus handler lookup to avoid linear searches

bdraco:
The time complexity in _find_message_handler was a bit much

There are also MANY typing fixes here since I added typing every place around the change that was missing to make sure I didn't make a mistake with the lookups


src/dbus_fast/message_bus.py
```diff
@@ -945,24 +932,18 @@ def _find_message_handler(
...
+            and (interfaces := self._path_exports.get(msg.path)) is not None
+            and (interface := interfaces.get(msg.interface)) is not None
+            and (
+                handler := ServiceInterface._get_enabled_handler_by_name_signature(
+                    interface, self, msg.member, msg.signature
+                )
+            )
+            is not None
+        ):
+            return handler
```
bdraco, line 946: All of the refactoring was so we could do this .. which should have a time complexity of O(3) instead of O(x)*O(y)*O(z)

src/dbus_fast/introspection.py
```diff
@@ -51,7 +51,7 @@ class Arg:
     def __init__(
         self,
         signature: Union[SignatureType, str],
-        direction: Optional[list[ArgDirection]] = None,
+        direction: Optional[ArgDirection] = None,
```
bdraco, line 54: this was just plain wrong

bdraco:
Whats missing coverage is from the all type ignores...

PR #409, opened 2025-03-06
fix: ensure proxy object tasks do not get garbage collected prematurely

### LINKED ISSUE #388: proxy_object can create tasks without holding a strong reference
They may be garbage collected before completed

https://github.com/Bluetooth-Devices/dbus-fast/blob/7f3552bf02603f4ad021e957576710a2b1846136/src/dbus_fast/proxy_object.py#L140

bdraco:
There was a task that did not have a strong reference held

fixes #388


src/dbus_fast/proxy_object.py
```diff
@@ -136,11 +139,16 @@ def _message_handler(self, msg: Message) -> None:
                 data = body
 
             cb_result = handler.fn(*data)
+            raise ValueError("signal handlers cannot return values")
```
bdraco, line 142: FAILED tests/client/test_signals.py::test_complex_signals - assert 0 == 1
FAILED tests/client/test_signals.py::test_varargs_callback - assert 0 == 1
FAILED tests/client/test_signals.py::test_kwargs_callback - assert 0 == 1

PR #421, opened 2025-03-13
fix: multiple calls on the root logger instead of module logger

bdraco:
detected by ruff 0.1.0 LOG015

### DIFF
--- a/src/dbus_fast/message_bus.py
+++ b/src/dbus_fast/message_bus.py
@@ -334,7 +334,7 @@ class BaseMessageBus:
                 try:
                     raise e
                 except Exception:
-                    logging.error(
+                    _LOGGER.error(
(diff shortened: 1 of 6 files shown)


src/dbus_fast/message_bus.py
```diff
@@ -1033,8 +1033,7 @@ def is_result_complete() -> bool:
 
         # first build up the result object to know when it's complete
         result: dict[str, dict[str, Any]] = {
-            node: {interface: None for interface in self._path_exports[node]}
-            for node in nodes
+            node: dict.fromkeys(self._path_exports[node]) for node in nodes
```
bdraco, line 1036:
```suggestion
            node: {interface: None for interface in self._path_exports[node]}
            for node in nodes
```

PR #424, opened 2025-03-17
fix: don't limit python upper bound

dlech:
This requires all projects that depend on dbus-fast to have to set the same limit, which is an unnecessary burden.


bdraco (approved):

dlech:
Some interesting reading on the subject: https://github.com/zanieb/poetry-relax?tab=readme-ov-file#references

I didn't do the same for other dependencies since they aren't runtime dependencies.

dlech:
Oops, forgot `poetry lock`.

bdraco:
> Oops, forgot `poetry lock`.

already on it

bdraco:
little bit of trial and error to find a working pygobject version since we are blocked updating because of the compile issues in https://github.com/Bluetooth-Devices/dbus-fast/pull/419

bdraco:
```

The current project's supported Python range (>=3.9) is not compatible with some of the required packages Python requirement:
  - pygobject requires Python <4.0,>=3.9, so it will not be installable for Python >=4.0


```

bdraco:
https://gitlab.gnome.org/GNOME/pygobject/-/blob/main/pyproject.toml?ref_type=heads#L25

dlech:
Meh, so we have to convince them to do the same and drop the upper limit.

dlech:
https://gitlab.gnome.org/GNOME/pygobject/-/merge_requests/418

bdraco (approved):

pyproject.toml
```diff
@@ -49,7 +49,7 @@ pytest = ">=7,<9"
 pytest-cov = ">=3,<7"
 pytest-asyncio = ">=0.19,<0.26"
 pycairo = "^1.21.0"
-PyGObject = "^3.42.2"
+PyGObject = {version = ">=3.50,<3.51", python = ">=3.9,<4"}
```
dlech, line 52:
```suggestion
PyGObject = {version = ">=3.50,<3.51", python = "<4"}
```

Does this work so we don't have to bump the Python version in 2 places?

bdraco: probably.. testing now

dlech: You are too fast. 😝

bdraco: https://github.com/Bluetooth-Devices/dbus-fast/compare/reduce_pin?expand=1

bdraco: https://github.com/Bluetooth-Devices/dbus-fast/pull/425

PR #425, opened 2025-03-17
chore: remove lower python bound from PyGObject

bdraco:


bdraco:
Works.. Now just need to figure out why > 3.50 is broken in https://github.com/Bluetooth-Devices/dbus-fast/pull/419

PR #446, opened 2025-04-03
fix: correct documentation URLs in README

joshuadavidthomas:
👋 Noticed a couple of the links to the docs were pointing to a non-existent project (or maybe it used to be there but was changed?).


dlech:
Not a link, but the same name is used in line 199 that should probably be changed too.

bdraco:
Thanks @joshuadavidthomas 

Should be good to go once dlech's comment is addressed

bdraco (approved):

PR #456, opened 2025-05-03
DNM: Prepare for Cython 3.1

bdraco:


bdraco:
Looks like we are good to go here

Issue #483, opened 2025-07-27
Unable to pass multiple file descriptors


radium226:
Hello! I think I've found a bug regarding file descriptors. I'm pretty sure it's related to something [here](https://github.com/Bluetooth-Devices/dbus-fast/blob/main/src/dbus_fast/_private/util.py) but I can't figure out. 

Thanks a lot for your help! I can create a PR, but I think I need guidance for this 😓 

**Describe the bug**
Let's say I have an `Foo` service interface with a `Bar(fd_1: "h", fd_2: "h")` method, the `fd_2` will always be `None` when I invoke the `call_bar`. If I try using `fds: "hh"` or `fds: "ah"`, it doesn't work either. 

**To Reproduce**
Just create a service interface that requires multiple file descriptors and try to call it. 

**Additional context**
n/a

PR #484, opened 2025-07-27
fix: passing multiple fds

radium226:
I think it fixes [this bug](https://github.com/Bluetooth-Devices/dbus-fast/issues/483)


bdraco:
I think you submitted this when Github was having issues. Merged in main to re-run CI

bdraco:
Its unclear if you have tested this change or not from the opening text.

Thanks

bdraco:
Fix looks correct for and likely a buffer underflow

PR #491, opened 2025-08-21
fix(message_bus): prevent socket leak on connection error

skbeh:
In the `_setup_socket` method, if an exception occurred during the connection attempt, the underlying socket file descriptor was not being closed.

This commit refactors the socket connection logic to ensure that if an error is present after trying all available bus addresses, the socket is explicitly closed before the connection error is raised.


bdraco:
Please add test coverage for this change. Thanks!

tests/test_message_bus.py
```diff
@@ -0,0 +1,31 @@
+import pytest
+
+from dbus_fast.errors import InvalidAddressError
+from dbus_fast.glib import MessageBus
+
+
```
dlech, line 6: It looks like these tests just test that an exception was raised. I don't seen anything that tests that resources were not leaked.

src/dbus_fast/message_bus.py
```diff
@@ -722,6 +722,13 @@ def _setup_socket(self) -> None:
                 raise InvalidAddressError(f"got unknown address transport: {transport}")
 
         if err:
+            self._fd = None
+            if self._stream:
+                self._stream.close()
+            self._stream = None
+            if self._sock:
+                self._sock.close()
+            self._sock = None
```
dlech, line 731: I think a `contextlib.ExitStack` would be a cleaner way to handle the cleaunup here. Just call `stack.pop_all()` to avoid the cleanup callbacks on a successful return.

Plus it would also handle other error cases where an exception is raised earlier.

dlech:
The coverage error is now only due to lines that were already missing before this PR. We've actually decreased the uncovered lines from 9 to 3.

It would be nice to get the mocks working, but it's not something that I can personally prioritize. So I would say this is good enough to merge now and if someone cares to, fix up the existing tests and add some more later.

PR #526, opened 2025-11-08
feat!: use tuple for STRUCT types instead of list

dlech:
Return tuple instead of list for STRUCT types when unmarshaling D-Bus messages. Marshalling still accepts both lists and tuples for STRUCT types.

This is a breaking change for anyone who was relying on STRUCT types being returned as lists. Operations like indexing and unpacking will still work the same, so hopefully the impact is minimal.

The motivation behind this is to allow proper Python type hints in `dbus_fast`. Python lists can only have a single type for all elements which is not compatible with D-Bus STRUCT types. Python tuples, on the other hand, exactly match the D-Bus STRUCT semantics.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -138,3 +138,6 @@ dmypy.json
 
 # Cython debug symbols
 cython_debug/
+
+# Cython generated files
+*.c
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -1,6 +1,6 @@
 [tool.poetry]
 name = "dbus-fast"
-version = "2.46.1"
+version = "2.46.4"
 description = "A faster version of dbus-next"
 authors = ["Bluetooth Devices Authors <bluetooth@koston.org>"]
 license = "MIT"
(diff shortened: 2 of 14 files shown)


dlech:
This is something that has bugged me for a long time. I never bothered to fix it though because it is a breaking change and typing in `dbus_next` isn't that great anyway so it didn't seem worth it. But now I have a plan to improve the type hints, so this become necessary for type hints to actually be correct.

I think the risk of breaking users is hopefully low, but I haven't gone looking for actual users to see how many people are using list methods on these values that would fail if called on a tuple. That would be a time consuming task. And if we do break anyone, it should be a fairly trivial fix since they just have to do `struct = list(struct)` to turn it back into a list. And this fixed code will be compatible with both 2.x and 3.x.

dlech:
Forgot to mention in the commit message that two of the test cases had to be pulled out of `messages.json` since they contain tuples and there isn't a way to represent that in json.

dlech:
> it seems in quite some cases we actually use `tuple` as type annotations for the struct cases, maybe it used to be tuples at one point?

No, it has always been `list` - it was inherited from `dbus-next`. But `tuple` is the natural way to write the type hints since `list` simply doesn't work for any STRUCT that has more than one member type. So I could see why someone would use `tuple` for the type hint even if it wasn't technically correct.

bdraco:
As soon as I finish testing https://github.com/home-assistant/core/pull/156695 and it merges I'll get this one released

PR #533, opened 2025-11-08
feat: add cython support for Cython 3.2+

### LINKED ISSUE #532: Change version of Cython in build system requires
In the build section Cython is `<3.1.0`

https://github.com/Bluetooth-Devices/dbus-fast/blob/04d3cf02055ec041236025dbeaa68951510cd8e8/pyproject.toml#L107-L109

Is it possible to make this track the same version as in `tool.poetry.group.dev.dependencies` ?
https://github.com/Bluetooth-Devices/dbus-fast/blob/04d3cf02055ec041236025dbeaa68951510cd8e8/pyproject.toml#L48-L58

The build of `dbus-fast` fails in Void Linux since it has `Cython 3.1.4`, so we need to [patch](https://github.com/void-linux/void-packages/blob/37d5299a4cac88df03ae55189ebd67e38dd145f3/srcpkgs/python3-dbus-fast/patches/cython-3.1.patch) the `pyproject.toml` to make it build. It would be nice to not have to do that.

kaistian:
Closes #532

### DIFF
--- a/src/dbus_fast/_private/marshaller.pxd
+++ b/src/dbus_fast/_private/marshaller.pxd
@@ -28,12 +28,12 @@ cdef class Marshaller:
     )
     cdef unsigned int _align(self, unsigned int n)
 
-    cpdef write_boolean(self, object boolean, SignatureType type_)
+    cpdef write_boolean(self, bint boolean, SignatureType type_)
 
     @cython.locals(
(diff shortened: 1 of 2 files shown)


bdraco (approved):
Thanks @kaistian

PR #534, opened 2025-11-08
chore: always wait for disconnect in async tests

dlech:
Add a wait for disconnect for a buses used in async tests to ensure that all resources are cleaned up before the test ends. This helps to catch more bugs in tests that would otherwise be overlooked.

The end goal here is to be able to change `-Wdefault` to `-Werror` in the pytest options. There are still a few bugs to fix before we can actually do that. So this is just an intermediate step to help us find and fix those bugs.


dlech:
This already helped find the bug fixed by #535. We can rebase this branch once that PR is merged and the CI failure will go away.

PR #535, opened 2025-11-08
fix: ordering in BaseMessageBus.export()

dlech:
Move the assignment of `self._path_exports[path][interface.name]` after the call to `ServiceInterface._add_bus()`. The latter can raise an exception, which would leave the former in a bad state.

The effect of this was that `bus.wait_for_disconnect()` would hang forever because `ServiceInterface._remove_bus()` would crash with a `KeyError`.


bdraco (approved):
Thanks @dlech

bdraco:
Will merge once I get the CI back in order

PR #538, opened 2025-11-16
fix: move call to close socket

dlech:
Move the call for closing the D-Bus socket from `dbus_fast.aio.MessageBus.disconnect()` to `dbus_fast.BaseMessageBus._finalize()`. The socket it opened by `BaseMessageBus`, so it should be closed by it as well. Also, `disconnect()` may not always be explicitly called, e.g. when connecting to the bus, so we need this to be in the `_finalize()` to ensure that it is always closed.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -1,6 +1,6 @@
 [tool.poetry]
 name = "dbus-fast"
-version = "2.46.1"
+version = "2.46.2"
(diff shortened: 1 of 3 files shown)


bdraco:
> Just did a merge to fix it since we squash anyway.

Yeah I prefer the merge actually since we squash and it makes reviews easier to follow

PR #540, opened 2025-11-16
chore: change pytest warning flag from default to errors

dlech:
This changes the pytest configuration in pyproject.toml to treat all Python warnings as errors. This will help catch potential issues like leaking resources or deprecated features during testing.

### DIFF
--- a/src/dbus_fast/_private/address.py
+++ b/src/dbus_fast/_private/address.py
@@ -1,3 +1,5 @@
+# cython: freethreading_compatible = True
+
 from __future__ import annotations
 
 import os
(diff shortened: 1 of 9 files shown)


bdraco:
Looks like we still have some close races

dlech:
Yup, more work to do.

dlech:
Heh, that fixed Cython and broke not-Cython. :hammer_and_wrench:

dlech:
Python 3.14 warnings were fixed by https://github.com/Bluetooth-Devices/dbus-fast/commit/9de4dbaf8c43eb5f4c1cf68d59f6ed1da80b7b81

dlech:
I managed to make Cython and not Cython happy at the same time.

Only thing left is how to handle no-GIL tests that are actually enabling GIL. For glib, I'm tempted to just ignore the warning. Not sure what the status of Cython and no-GIL is though.

PR #548, opened 2025-11-16
feat: change aio MessageBus.call return type to Message

dlech:
Change the return type of `dbus_fast.aio.MessageBus.call` from `Message | None` to `Message` and deprecate its use for messages that do not expect a reply.

This method is primarily intended for method calls, which always expect a reply. When using strict type checking, this makes for unnecessary work to handle the `None` case when calling methods.

In those cases, the `call()` method just calls `send()` internally anyway, so it would make more sense for users to call `send()` directly instead. Therefore, the deprecation notice recommends this. We just issue a `DeprecationWarning` for now, but in a future major release we may remove the ability to use `call()` for messages that do not expect a reply.

### DIFF
--- a/tests/test_aio_low_level.py
+++ b/tests/test_aio_low_level.py
@@ -125,6 +125,11 @@ async def test_sending_messages_between_buses():
 
     msg.serial = bus2.next_serial()
     msg.flags = MessageFlag.NO_REPLY_EXPECTED
+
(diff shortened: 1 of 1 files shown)


bdraco (approved):

bdraco:
Tests are failing, however I think this is a good design change once the CI is passing

PR #551, opened 2025-11-17
feat: allow returning multiple values as tuples instead of lists

dlech:
In Python, the natural way to return multiple values from a function is
as a tuple. Change the docs to recommend returning tuples instead of lists
for D-Bus methods and signals that return multiple values.

Tests are updated as well to show that tuples are preferred (and that
this is not a breaking change).


dlech:
I'll have to think about this some more.

There are two sides to this, only one of which has been addressed so far.

The first is relaxing the allowable types for `body` as an input. This is what has been done here so far. Other than the one type change for Cython that is needed to make it work at runtime, it is just a matter of type hints. I've made everything `list[Any] | tuple[Any, ...]`, but that is rather verbose and we could actually replace it with `Sequence[Any]` (from `collections.abc`).

The part I haven't addressed though is that I've just realized that in addition to #526 there is also the case where returning multiple values should also be implemented as a `tuple` to allow for correct type hints. However, it is currently implemented as a `list`. Where this matters is for proxy types. We likely need it to be a list when unmarshalling messages to avoid issues with tuple immutability, i.e. a performance cost for having to create a new tuple object every time we unmarshal the next element.

PR #560, opened 2025-11-19
feat: added aio MessageBus 'loop' optional parameter

depaolim:
Added the optional loop parameter to asyncio's MessageBus object. If not specified, it uses the running_loop parameter as before. This can be useful for special cases where the programmer needs to execute operations in a specific loop of their choosing.

### DIFF
--- a/src/dbus_fast/aio/message_bus.py
+++ b/src/dbus_fast/aio/message_bus.py
@@ -207,6 +207,11 @@ class MessageBus(BaseMessageBus):
         negotiate_unix_fd: bool = False,
         loop: asyncio.AbstractEventLoop | None = None,
     ) -> None:
+        """
+        .. versionchanged:: 3.2.0
+            Added a new optional parameter "loop".
+            Use this only for special cases (e.g., multithreading, where each thread needs its own loop).
+        """
         super().__init__(bus_address, bus_type, ProxyObject, negotiate_unix_fd)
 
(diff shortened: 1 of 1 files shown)


depaolim:
Here I am, finally back. I’m submitting a few points of reflection; in any case, I really appreciate this repository, so I’ll trust your judgment on whether introducing this parameter is appropriate.

1. Without the `loop` parameter, `MessageBus` would attempt to call `asyncio.get_running_loop()`, which may either fail or return the wrong loop in multi-threaded contexts where each thread manages its own event loop. In short—as the Zen of Python reminds us—explicit is better than implicit. :-)

2. According to the documentation for `asyncio.run` (see: https://docs.python.org/3/library/asyncio-runner.html#asyncio.run), it “_Execute coro in an asyncio event loop and return the result. [..] managing the asyncio event loop, finalizing asynchronous generators, and closing the executor._”
However, automatically closing the executor may not always be what the programmer intends.

3. The pattern of explicitly passing a `loop` argument is already used within `asyncio` itself. For example, see `asyncio.StreamReader.__init__`:
https://github.com/python/cpython/blob/202fce0dbde1da32d8abc2eb59ddfce6f6a3c9fa/Lib/asyncio/streams.py#L422

As I mentioned, I trust your judgment. I agree that the parameter isn’t strictly essential, so I’m also fine with closing the PR without merging.

PR #562, opened 2025-11-27
feat: add basic support for Annotated[...] DBus signature annotations

Sergi030:
This PR adds support for using PEP 593 Annotated types to define DBus signature annotations in dbus-fast.
https://peps.python.org/pep-0593/
https://typing.python.org/en/latest/spec/qualifiers.html#annotated

This allows combining Python type hints and DBus signatures in a single annotation, like the following one:

```bash
from typing import Annotated

def Version(self) -> Annotated[str, "s"]:
    return "1.0.0"
```

This change is mainly intended to support modern Python typing while preserving the existing dbus-fast string-based DBus signature notation.

This pr should be fully backwards-compatible, the existing code using plain string signatures continues to behave exactly as before.

### DIFF
--- a/src/dbus_fast/_private/util.py
+++ b/src/dbus_fast/_private/util.py
@@ -120,7 +120,8 @@ def parse_annotation(annotation: str) -> str:
     if get_origin(annotation) == Annotated:
         try:
             annotation = get_args(annotation)[1]
-        except IndexError:
+        except IndexError:  # pragma: no cover
+            # should not be possible due Annotated always having >=1 args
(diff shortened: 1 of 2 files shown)


Sergi030:
Thanks for taking the time to look at and review this, @bdraco and @dlech. Since @dlech already has another PR #563   with documentation updates and a related discussion, I’m going to withdraw this PR.

Also, my apologies for pushing this feature again and potentially rushing things.

Thanks again for the feedback and for your work on the project!

PR #563, opened 2025-11-28
feat!: annotated types for D-Bus signatures

dlech:
This is an alternative to #562.

I did not submit this yet because I got stuck on #551. But since there is another PR for the same thing, I'm posting it now to be able to compare.

Compared to #562 the user-facing change is the same but the implementation is different.

Differences to #562:
* The check for `Annotated` is a bit simpler. We just check for the `__metadata__` attribute.
* Includes documentation.
* Adds a `dbus_fast.annotations` module with type aliases for common types.
* Converts all tests to use annotated types.
* Lacks tests for checking error paths.

### DIFF
--- a/docs/source/high-level-service/index.rst
+++ b/docs/source/high-level-service/index.rst
@@ -78,7 +78,7 @@ constructor to use unix file descriptors.
 
 .. code-block:: python3
 
-    from dbus_fast.annotations import (DBusSignature, DBusBool, DBusByte, DBusInt,
+    from dbus_fast.annotations import (DBusSignature, DBusBool, DBusByte, DBusInt32,
                                        DBusStr)
     from dbus_fast.aio import MessageBus
     from dbus_fast.service import (ServiceInterface,
@@ -86,6 +86,7 @@ constructor to use unix file descriptors.
     from dbus_fast import Variant, DBusError
 
     import asyncio
+    from typing import Annotated
 
     FrobateReturnDBusType = Annotated[dict[int, str], DBusSignature("a{us}")]
     BazifyBarDBusType = Annotated[tuple[int, int, int], DBusSignature("(iiu)")]
@@ -98,7 +99,7 @@ constructor to use unix file descriptors.
             self._bar = 105
 
         @dbus_method()
-        def Frobate(self, foo: DBusInt, bar: DBusStr) -> FrobateReturnDBusType:
+        def Frobate(self, foo: DBusInt32, bar: DBusStr) -> FrobateReturnDBusType:
             print(f'called Frobate with foo={foo} and bar={bar}')
 
             return {
--- a/docs/source/high-level-service/index.rst
+++ b/docs/source/high-level-service/index.rst
@@ -78,6 +78,8 @@ constructor to use unix file descriptors.
 
 .. code-block:: python3
 
+    from dbus_fast.annotations import (DBusSignature, DBusBool, DBusByte, DBusInt,
+                                       DBusStr)
     from dbus_fast.aio import MessageBus
     from dbus_fast.service import (ServiceInterface,
                                    dbus_method, dbus_property, dbus_signal)
@@ -85,13 +87,18 @@ constructor to use unix file descriptors.
 
     import asyncio
 
+    FrobateReturnDBusType = Annotated[dict[int, str], DBusSignature("a{us}")]
+    BazifyBarDBusType = Annotated[tuple[int, int, int], DBusSignature("(iiu)")]
+    BazifyReturnDBusType = Annotated[tuple[Variant, Variant], DBusSignature("vv")]
+    MorgifyBarDBusType = Annotated[tuple[int, int, list[Variant]], DBusSignature("(iiav)")]
+
     class ExampleInterface(ServiceInterface):
         def __init__(self):
             super().__init__('com.example.SampleInterface0')
             self._bar = 105
 
         @dbus_method()
-        def Frobate(self, foo: 'i', bar: 's') -> 'a{us}':
+        def Frobate(self, foo: DBusInt, bar: DBusStr) -> FrobateReturnDBusType:
             print(f'called Frobate with foo={foo} and bar={bar}')
 
             return {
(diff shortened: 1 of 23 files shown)


dlech:
As expected, the Cython tests are failing with:

```
FAILED tests/client/test_methods.py::test_aio_proxy_object - dbus_fast.errors.DBusError: The service interface raised an error: Argument 'body' has incorrect type (expected list, got tuple).
```

This is the problem that #551 is trying to solve.

To illustrate, the problem, consider that it is possible for D-Bus methods to return multiple values.

In Python, the natural way to do this is with a tuple.

```python
    @dbus_method()
    def echo_example(self, what1: DBusInt, what2: DBusStr) -> Annotated[tuple[int, str], "is"]:
        return what1, what2
```

However, D-Bus fast, expects returning with a list, not a tuple. When not using Cython, a tuple still works because we don't internally use any list methods and only call APIs common between `list` and `tuple`. But Cython specifically checks for a `list` and fails.

The best we could do with typing for this would be:

```python
    @dbus_method()
    def echo_example(self, what1: DBusInt, what2: DBusStr) -> Annotated[tuple[int, str], "is"]:
        return [what1, what2]  # type: ignore
```

or 


```python
    @dbus_method()
    def echo_example(self, what1: DBusInt, what2: DBusStr) -> Annotated[list[Any], "is"]:
        return [what1, what2]
```

Neither of which are idea.

So this is why I was wanting to try to solve #551 first before attempting supporting annotated types.

src/dbus_fast/_private/util.py
```diff
@@ -115,6 +115,13 @@ def raise_value_error() -> None:
 
     if not annotation or annotation is inspect.Signature.empty:
         return ""
+
+    # If there is __metadata__, it means this is typing.Annotated. In that case,
+    # the first item in __metadata__ should contain the D-Bus signature string.
+    metadata = getattr(annotation, "__metadata__", None)
+    if metadata is not None:
+        annotation = metadata[0]
```
dlech: What @Sergi030 did in #562 is a bit more robust compared to what I did here. 😄

dlech: Thanks for bringing this up. I'm not really familiar with how `Annotated` is used in the wild, so it's helpful and interesting to hear about it.

It sounds like a good idea. My only objection is that it makes it even more verbose than it already is. The methods these will be used on are not really regular Python methods, so the chances of needing to add additional annotations seems low, but what do I know? :smile:

So I could go either way on it.

christian-hnz: Here are a few examples [fastapi](https://fastapi.tiangolo.com/tutorial/body-multiple-params/), [pydantic](https://docs.pydantic.dev/latest/concepts/validators/#field-validators), [typer](https://typer.tiangolo.com/tutorial/parameter-types/number/?h=annotated#number) ... in those cases, there are good reasons for the wrapper.

There was also a PEP to have a std. way of including docs in the metadata [like this](https://typing-extensions.readthedocs.io/en/latest/#annotation-metadata) but it was withdraw ... after being criticized for adding too much *verbosity* :).

On the other hand, the wrapper may play a clarifying role: `Annotated[int, DBusSignature("i")]` or just `Annotated[int, Sig("i")]` makes it pretty evident what the string is good for. But other than that I don't have a usecase. It's just a matter of taste ...

bdraco (approved):

bdraco:
Looks like it was just the `__name__` call that broke because we used a partial in the tests.

Need to check all the other ecosystem scanners to make sure they don't use partials

bdraco:
Looks good. going to ask the bot to see if there is anything I missed

docs/source/high-level-service/index.rst
```diff
@@ -78,20 +78,27 @@ constructor to use unix file descriptors.
...
+    MorgifyBarDBusType = Annotated[tuple[int, int, list[Variant]], DBusSignature("(iiav)")]
+
     class ExampleInterface(ServiceInterface):
         def __init__(self):
             super().__init__('com.example.SampleInterface0')
             self._bar = 105
 
         @dbus_method()
-        def Frobate(self, foo: 'i', bar: 's') -> 'a{us}':
+        def Frobate(self, foo: DBusInt, bar: DBusStr) -> FrobateReturnDBusType:
```
dlech:
```suggestion
        def Frobate(self, foo: DBusInt32, bar: DBusStr) -> FrobateReturnDBusType:
```

src/dbus_fast/glib/proxy_object.py
```diff
@@ -56,8 +56,8 @@ def callback(error: Exception, result: list(Any)):
 
     To *synchronously* call a method, use the ``call_[METHOD]_sync()`` form.
     The ``result`` corresponds to the *out arg* of the introspection method
-    definition. If the method has more than one otu arg, they are returned
-    within a :class:`list`.
+    definition. If the method has more than one out arg, they are returned
+    within a :class:`tuple`.
```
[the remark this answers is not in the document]
dlech: Saving cleaning this up for #551.

src/dbus_fast/service.py
```diff
@@ -117,7 +118,27 @@ def echo(self, val: 's') -> 's':
...
+        from dbus_fast.annotations import DBusStr, DBusUInt32
+
+        @dbus_method()
+        def echo(self, val: DBusStr) -> DBusStr:
+            return val
+
+        @dbus_method()
+        def echo_two(
+            self, val1: DBusStr, val2: DBusUInt32
+        ) -> Annotated[tuple[str, int], "su"]:
```
dlech:
```suggestion
        ) -> Annotated[tuple[str, int], DBusSignature("su")]:
```

src/dbus_fast/service.py
```diff
@@ -206,7 +228,27 @@ def string_signal(self, val) -> 's':
...
+        from dbus_fast.annotations import DBusStr
+
+        @dbus_signal()
+        def string_signal(self, val: str) -> DBusStr:
+            return val
+
+        @dbus_signal()
+        def two_strings_signal(
+            self, val1: str, val2: str
+        ) -> Annotated[tuple[str, str], "ss"]:
```
dlech:
```suggestion
        ) -> Annotated[tuple[str, str], DBusSignature("ss")]:
```

docs/source/high-level-service/index.rst
```diff
@@ -78,20 +78,27 @@ constructor to use unix file descriptors.
...
 .. code-block:: python3
 
+    from dbus_fast.annotations import (DBusSignature, DBusBool, DBusByte, DBusInt,
+                                       DBusStr)
     from dbus_fast.aio import MessageBus
     from dbus_fast.service import (ServiceInterface,
                                    dbus_method, dbus_property, dbus_signal)
     from dbus_fast import Variant, DBusError
 
     import asyncio
```
dlech:
```suggestion
    import asyncio
    from typing import Annotated
```

PR #570, opened 2025-12-02
fix!: defer socket.connect() from __init__ to connect()

### LINKED ISSUE #569: Blocking I/O in MessageBus init
**Describe the bug**
blockbuster found that this library is doing blocking I/O in the MessageBus constructor:

```
  File "/usr/src/supervisor/supervisor/dbus/manager.py", line 125, in load
    self._bus = connected_bus = await MessageBus(
                                      ~~~~~~~~~~^
        bus_type=BusType.SYSTEM
        ^^^^^^^^^^^^^^^^^^^^^^^
    ).connect()
    ^
  File "/usr/local/lib/python3.13/site-packages/dbus_fast/aio/message_bus.py", line 209, in __init__
    super().__init__(bus_address, bus_type, ProxyObject, negotiate_unix_fd)
    ~~~~~~~~~~~~~~~~^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "src/dbus_fast/message_bus.py", line 176, in dbus_fast.message_bus.BaseMessageBus.__init__
  File "src/dbus_fast/message_bus.py", line 747, in dbus_fast.message_bus.BaseMessageBus._setup_socket
  File "src/dbus_fast/message_bus.py", line 709, in dbus_fast.message_bus.BaseMessageBus._setup_socket
  File "/usr/local/lib/python3.13/site-packages/blockbuster/blockbuster.py", line 109, in wrapper
    raise BlockingError(func_name)
blockbuster.blockbuster.BlockingError: Blocking call to socket.socket.connect

```

It seems the library is calling `socket.connect()` in `_setup_socket`.

**To Reproduce**
Connect to D-Bus with blockbuster enabled.

**Additional context**

agners:
Fixes #569.

## Summary

The aio `MessageBus` constructor used to perform a blocking `socket.connect()` inside `BaseMessageBus.__init__()` (via `_setup_socket`). That violates async hygiene and trips blocking-call detectors like blockbuster in async contexts (notably Home Assistant Supervisor).

This PR moves socket creation and connection for the **aio** bus into `MessageBus.connect()`, so `MessageBus()` is now side-effect-free and connection errors surface from `connect()` instead of construction. The glib bus is intentionally left with its blocking-in-`__init__` behavior — only the shared helper has been factored out — because we don't know who relies on the existing glib semantics and didn't want to break them on a hunch.

## BREAKING CHANGE (aio only)

Connection errors that previously came from `aio.MessageBus(...)` instantiation are now raised from `connect()`. Code using the common pattern is unaffected:

```python
bus = await MessageBus(bus_type=BusType.SYSTEM).connect()  # still works, exceptions just come from .connect()
```

Code that wraps the constructor itself in `try/except` needs to move the `try` to cover `connect()`. The glib bus is unchanged.

## Approach

- **Shared helper:** `BaseMessageBus._create_socket_for_transport(transport, options)` (`@staticmethod`) creates an unconnected socket+stream and returns the address. Used by both `aio.MessageBus.connect()` and `glib.MessageBus.__init__()`. Options are parsed before any resources are allocated so malformed addresses fail without leaking.
- **aio `connect()`:** iterates transports with `await loop.sock_connect()`, uses `ExitStack` per attempt so a `CancelledError` doesn't leak the freshly-created sock/stream. The unix→tcp fallback is preserved; on total failure the last error is raised (a `REVISIT:` comment notes we could surface all of them via `ExceptionGroup` once Python 3.10 is dropped).
- **aio `connect()` post-connect cleanup:** writer creation, `_authenticate()`, and reader registration are wrapped in `try/except BaseException` that closes the socket/stream and clears `_sock`/`_stream`/`_fd`/`_writer` on failure, so an auth error no longer leaks the socket.
- **aio `_writer`:** constructed inside `connect()` after the socket is set, instead of being created in `__init__` with `None` references and back-patched. `__init__` just sets `self._writer = None`.
- **glib:** intentionally unchanged behavior. `__init__` inlines the per-transport setup loop using the same shared helper. No `_setup_socket` method on the base class anymore.
- **Docs:** `.. versionchanged:: v5.0.0` notes on `aio.MessageBus.connect()` and the class docstring. `_create_socket_for_transport` documents both `InvalidAddressError` and `ValueError` (for non-integer `tcp:port=...`).

## Relationship to #592

#592 (`async with bus` context manager) is easier on top of this PR — `__aenter__` no longer has to work around the constructor connecting the socket. dlech confirmed in #592 that this should land first.

## Tests

- `test_aio_connect_falls_back_between_transports` — multi-transport address, all fail, last error raised.
- `test_aio_connect_cleanup_after_socket_connected` — auth failure after socket is connected clears all bus state.
- `test_aio_connect_hello_error_propagates_and_clears_state` — error reply to the initial Hello tears down sock/stream/fd/writer.
- `test_create_socket_for_transport_{unix,tcp}_makefile_failure_closes_socket` — verifies the makefile-failure cleanup path.
- `test_glib_low_level` gets a transport-fallback test covering the glib path.
- `test_send_reply` `MockBus` updated to populate `_sock`/`_stream` in `connect()` rather than `__init__`, matching the new pattern.

Fixes blocking-I/O detection errors like:

```
blockbuster.BlockingError: Blocking call to socket.socket.connect
```

Fixes: #569


dlech (commented):
This isn't the worst breaking change ever since it is mostly only about where the exception is raised.

I'm not sure I like the suggestion of adding a paramter to opt into the new behavior. The new behavior is the more "correct" behavoir, we it would be best if we just did the right thing without having to depend on the user knowing what is best. We could add a deprecation warning to help users find this out if they don't set the option. But then we are stuck with the option forever, or we have to have another deprecation period and another breaking change to remove it.

Maybe we could write a quick script to check libraries that depend on dbus-fast to see if anyone is actually catching the exception from the constructor. If there aren't any, maybe we don't worry about it. Or if there are just a few, we could notify them of the change.

src/dbus_fast/aio/message_bus.py
```diff
@@ -232,6 +232,12 @@ async def connect(self) -> MessageBus:
               the DBus daemon failed.
             - :class:`Exception` - If there was a connection error.
         """
+        try:
```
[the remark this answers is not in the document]
agners: Technically its a breaking change, so I went with 5.0.0, let me know if you think otherwise.

src/dbus_fast/message_bus.py
```diff
@@ -726,25 +732,31 @@ def _setup_socket(self) -> None:
...
+                    return
 
-                else:
-                    raise InvalidAddressError(
-                        f"got unknown address transport: {transport}"
-                    )
+                raise InvalidAddressError(f"got unknown address transport: {transport}")
+
+        # Should not normally happen, but just in case
+        raise TypeError("empty list of bus addresses given")  # pragma: no cover
```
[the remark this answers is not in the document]
agners: Part of async `connect()` now.

src/dbus_fast/message_bus.py
```diff
@@ -680,65 +678,70 @@ def _introspect_export_path(self, path: str) -> intr.Node:
...
-                        filename = options["path"]
-                    elif "abstract" in options:
-                        filename = b"\0" + options["abstract"].encode()
-                    else:
-                        raise InvalidAddressError(
-                            "got unix transport with unknown path specifier"
-                        )
+            except BaseException:
+                sock.close()
+                raise
```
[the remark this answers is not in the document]
agners: Seems legitimate, addressed with the latest commit.

src/dbus_fast/message_bus.py
```diff
@@ -680,65 +678,70 @@ def _introspect_export_path(self, path: str) -> intr.Node:
...
-                        return
+        if transport == "tcp":
+            sock = socket.socket(socket.AF_INET, socket.SOCK_STREAM)
+            try:
+                stream = sock.makefile("rwb")
+            except BaseException:
+                sock.close()
+                raise
+            ip_addr = options.get("host", "")
+            ip_port = int(options["port"]) if "port" in options else 0
```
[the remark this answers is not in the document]
agners: Same, addressed with the latest commit.

tests/test_send_reply.py
```diff
@@ -32,18 +31,18 @@ def close(self) -> None:
             pass
 
     class MockBus(BaseMessageBus):
+        def __init__(self) -> None:
+            super().__init__()
+            self._sock = MockClosable()  # type: ignore
+            self._stream = MockClosable()  # type: ignore
```
dlech, line 37: Should this be done in `connect()` now instead of `__init__()` to match the change this PR is making??

tests/test_send_reply.py
```diff
@@ -71,18 +70,18 @@ def close(self) -> None:
...
```
dlech: Should this be done in `connect()` now instead of `__init__()` to match the change this PR is making??

tests/test_message_bus.py
```diff
@@ -2,90 +2,126 @@
...
+
+def test_setup_socket_falls_back_between_transports() -> None:
+    """The sync _setup_socket() path (used by glib) iterates transports and
+    raises the last error if all fail.
+    """
+
+    class _Bus(BaseMessageBus):
+        pass
+
+    bus = _Bus.__new__(_Bus)
```
dlech: Do we really need to do `__new__()` and `__init__()` here instead of calling the constructor normally?

If yes, add a comment in the code to explain.

tests/test_message_bus.py
```diff
@@ -2,90 +2,126 @@
...
+    BaseMessageBus.__init__(
+        bus,
+        "unix:path=/there-is-no-way-that-this-file-should-exist;"
+        "tcp:host=127.0.0.1,port=1",
+    )
+
+    with pytest.raises(ConnectionRefusedError):
+        bus._setup_socket()
```
dlech: Why are we testing a "private" method?

src/dbus_fast/aio/message_bus.py
```diff
@@ -230,8 +235,50 @@ async def connect(self) -> MessageBus:
         :raises:
             - :class:`AuthError <dbus_fast.AuthError>` - If authorization to \
               the DBus daemon failed.
-            - :class:`Exception` - If there was a connection error.
+            - :class:`OSError` - If the socket could not be connected (e.g. \
+              :class:`ConnectionRefusedError`, :class:`FileNotFoundError`). \
+              When multiple transports are given, the error from the last \
+              attempted transport is raised.
```
dlech, line 250: When we drop support for Python 3.10, we could probably change this to an exception group to capture all exceptions.

Maybe worth a `REVIST: ` comment in the code.

src/dbus_fast/message_bus.py
```diff
@@ -680,65 +678,70 @@ def _introspect_export_path(self, path: str) -> intr.Node:
...
-            ip_port = 0
-
-            with ExitStack() as stack:
-                if transport == "unix":
-                    self._sock = stack.enter_context(
-                        socket.socket(socket.AF_UNIX, socket.SOCK_STREAM)
-                    )
-                    self._stream = stack.enter_context(self._sock.makefile("rwb"))
-                    self._fd = self._sock.fileno()
+    def _create_socket_for_transport(
```
dlech, line 680: This can be `@staticmethod` since it doesn't use `self`.

src/dbus_fast/message_bus.py
```diff
@@ -680,65 +678,70 @@ def _introspect_export_path(self, path: str) -> intr.Node:
...
-                    )
-                    self._stream = stack.enter_context(self._sock.makefile("rwb"))
-                    self._fd = self._sock.fileno()
+        raise InvalidAddressError(f"got unknown address transport: {transport}")
 
-                    if "host" in options:
-                        ip_addr = options["host"]
-                    if "port" in options:
-                        ip_port = int(options["port"])
+    def _setup_socket(self) -> None:
```
dlech: Can we just put this inline in the glib `connect()` method now like we did with the aio implementation?

src/dbus_fast/aio/message_bus.py
```diff
@@ -230,8 +235,50 @@ async def connect(self) -> MessageBus:
...
+            from ``MessageBus()`` are now raised from this method. This avoids
+            performing blocking I/O on the event loop during ``__init__``.
         """
+        last_err: Exception | None = None
+
+        for transport, options in self._bus_address:
+            sock, stream, address = self._create_socket_for_transport(
+                transport, options
+            )
+            sock.setblocking(False)
```
dlech: If this raises an exception, the `sock` and `stream` won't be closed.

Maybe not a big deal, since there isn't really a way to recover from that error. But we could use the `ExitStack` like the old implementation to fix that.

src/dbus_fast/aio/message_bus.py
```diff
@@ -230,8 +235,50 @@ async def connect(self) -> MessageBus:
...
+                stream.close()
+                sock.close()
+                last_err = e
+                continue
+
+            self._sock = sock
+            self._stream = stream
+            self._fd = sock.fileno()
+            self._writer.sock = sock
+            self._writer.fd = self._fd
```
dlech: This looks fragile.

I would either defer `self._writer = _MessageWriter(self)` to here or modify `_MessageWriter` to lazy-get these from `_MessageWriter.bus`.

Or add some comments both here and in `_MessageWriter.__init__()` to explain what is going on.

dlech (commented):
I think this looks in pretty good shape now. Just a few more comments...

src/dbus_fast/message_bus.py
```diff
@@ -680,65 +678,70 @@ def _introspect_export_path(self, path: str) -> intr.Node:
...
+            try:
+                stream = sock.makefile("rwb")
+            except BaseException:
+                sock.close()
+                raise
+            return sock, stream, address
+
+        if transport == "tcp":
+            ip_addr = options.get("host", "")
+            ip_port = int(options["port"]) if "port" in options else 0
```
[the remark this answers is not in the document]
dlech: not sure I agree with this suggestion

src/dbus_fast/aio/message_bus.py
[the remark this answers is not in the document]
dlech: This looks valid.

tests/test_message_bus.py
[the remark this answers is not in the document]
dlech: I already made the comment that it doesn't make sense to be testing this "private" method in the first place.

bdraco:
@bluetoothbot review

src/dbus_fast/aio/message_bus.py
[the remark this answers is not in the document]
bdraco: Addressed in eb70732: wrapped the post-connect setup in try/except that closes `sock`/`stream` and resets the bus state before re-raising.

tests/test_message_bus.py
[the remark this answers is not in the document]
bdraco: Addressed in eb70732 by removing the test entirely — the aio fallback test already covers the public connect() contract, and there's no value in poking the private `_setup_socket()`.

tests/test_message_bus.py
bluetoothbot: Agreed — `BaseMessageBus.__init__` accepts `bus_address=None` and `ProxyObject=None` defaults, so `_Bus("unix:...;tcp:...")` works directly without the `__new__()` workaround. The current form risks subtly diverging from real construction (e.g., skipping any future `__init_subclass__` hooks or violating `__slots__` invariants). Recommend just calling `_Bus("unix:...;tcp:...")` instead — and if there's a reason the constructor truly can't be used, add a comment explaining why.

tests/test_message_bus.py
bluetoothbot: Concur. The fallback behavior for the aio path is already covered by `test_aio_connect_falls_back_between_transports` through the public API. The only thing this test uniquely exercises is the *sync* fallback path used by glib, but coupling tests to a `_`-prefixed method ties them to implementation details. If you want explicit coverage for the glib path, prefer exercising it through `glib.MessageBus(...)` (skipping when glib/PyGObject isn't importable). If that's not viable in the test environment, the test could simply be dropped — `test_setup_socket_falls_back_between_transports` is largely structural duplication of the aio test.

src/dbus_fast/aio/message_bus.py
bluetoothbot: Worth fixing — it's actually a real leak in one case. `except Exception` does not catch `asyncio.CancelledError` (it derives from `BaseException` since Python 3.8), so if the task running `connect()` is cancelled while `await self._loop.sock_connect(sock, address)` is in flight, `CancelledError` propagates out and the freshly-created `sock`/`stream` are leaked until GC closes them (and `socket.makefile()` keeps a refcount on the socket, so the FD stays open until the stream is also GC'd). The `ExitStack` pattern the sync `_setup_socket` still uses handles this cleanly — push `stream.close` / `sock.close` as callbacks, call `stack.pop_all()` only after the `self._sock`/`self._stream`/`self._writer` assignments succeed.

src/dbus_fast/aio/message_bus.py
bluetoothbot: Strong agree. The smallest fix is option 1: defer `self._writer = _MessageWriter(self)` from `__init__` to `connect()` (right after the socket assignments), so the writer is constructed exactly once with a valid socket and no back-patching is needed. Lazy resolution via `self.bus._sock`/`self.bus._fd` (or `@property` accessors) is the more defensive option but means an extra attribute lookup on every send, which matters here since `write_callback` is on the hot path.

src/dbus_fast/message_bus.py
bluetoothbot: Agreed — `_create_socket_for_transport` reads no instance state. Marking it `@staticmethod` documents that fact explicitly and lets it be called without an instance (handy from any future helper or test). The Cython side stays unaffected since it's only annotated on the Python class, not in `.pxd`.

src/dbus_fast/message_bus.py
bluetoothbot: Plausible cleanup. With aio's path moved out, `_setup_socket` has exactly one remaining caller (`glib.MessageBus.__init__`). Inlining the small loop there would (a) let `_setup_socket` revert from `cpdef` back to `cdef` (or be removed entirely, dropping the `.pxd` churn), and (b) make it obvious in code review that the sync blocking-connect semantics are glib-specific rather than shared base-class behavior.

src/dbus_fast/aio/message_bus.py
bdraco: Addressed in 03232eb: switched the per-transport loop in `aio.connect()` to `ExitStack` so a `CancelledError` during `sock_connect` no longer leaks the sock/stream. Also changed the post-connect cleanup to `except BaseException` for the same reason.

src/dbus_fast/aio/message_bus.py
bdraco: Addressed in 03232eb: `self._writer = _MessageWriter(self)` is now created in `connect()` after the socket is assigned. No more back-patching of `_writer.sock`/`_writer.fd`.

src/dbus_fast/message_bus.py
bdraco: Addressed in 03232eb: `_setup_socket` is removed from `BaseMessageBus` and the loop is inlined into `glib.MessageBus.__init__`. `.pxd` reverted to its pre-PR state.

src/dbus_fast/message_bus.py
bdraco: Addressed in 03232eb: made `@staticmethod`.

bdraco:
@bluetoothbot review

bdraco:
Was hoping to get this one through but had to pause to fix security issues

bdraco:
@bluetoothbot rr

bdraco:
@bluetoothbot review

bdraco:
I think its in good shape now but need to reconcile all the review comments

tests/test_send_reply.py
agners: For the test purpose it doesn't really matter, and for glib we still set this in `__init__()`. But yeah aio is what we care more about, so probably better to align with it :+1:

src/dbus_fast/message_bus.py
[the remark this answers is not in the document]
agners: `ValueError` seems fine to me, too. I've added it to the docstring.

src/dbus_fast/aio/message_bus.py
agners: Adressed in 03232eb.

tests/test_message_bus.py
agners: Adressed in 03232eb.

tests/test_message_bus.py
agners: Adressed with 538833a.

tests/test_send_reply.py
agners: Adressed in 0faba1c.

agners:
I've went through all the review comments and double checked if they are addressed/addressed missing ones. From my point of view the PR is ready for another round of review.

tests/test_message_bus.py
```diff
@@ -1,99 +1,237 @@
...
+
+        assert bus._sock is None
+        assert bus._stream is None
+        assert bus._fd is None
+        assert bus._writer is None
+    finally:
+        listener.close()
+
+
+def _assert_makefile_failure_closes(
```
dlech: It looks like this could be turned into a parameterized test instead of two tests that call a helper function.

dlech (commented):
I didn't have time to really go through the cleanup logic again, but given there are enough tests for it now, I think this is good now.

Just one comment on making tests more readable.

bdraco:
Quick review on my laptop and the code looks right. Its a bit hard to follow so I want to do a deep 2 screen check on my desktop before approving.  Than have AI check to make sure we didn't miss any review comments.  I'll also ask the bot to do another pass.

bdraco:
@bluetoothbot review

bdraco (approved):

bdraco:
Thanks @agners @dlech

src/dbus_fast/message_bus.py
```diff
@@ -707,7 +708,12 @@ def _create_socket_for_transport(
 
         if transport == "tcp":
             ip_addr = options.get("host", "")
-            ip_port = int(options["port"]) if "port" in options else 0
+            try:
+                ip_port = int(options["port"]) if "port" in options else 0
```
dlech, line 712: Not sure why this isn't `ip_port = int(options.get("port", 0))`.

PR #623, opened 2026-05-02
build: drop stale hash comments from dependabot managed actions uses

eshattow:
Drops the trailing line comment with version. These comments are stale since this is managed by dependabot.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -14,8 +14,8 @@ jobs:
   lint:
     runs-on: ubuntu-latest
     steps:
-      - uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # v4
-      - uses: actions/setup-python@a309ff8b426b58ec0e2a45f0f869d46889d02405 # v5
+      - uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd
+      - uses: actions/setup-python@a309ff8b426b58ec0e2a45f0f869d46889d02405
         with:
           python-version: "3.10"
       - uses: pre-commit/action@2c7b3805fd2a0fd8c1884dcaebf91fc102a13ecd # v3.0.1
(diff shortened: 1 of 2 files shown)


dlech:
stale comment commit should be in a separate PR (we always do squash and merge in this repo and would be good to keep that separate)

dlech:
`Detached HEAD state cannot match any release groups; no release will be made`

is caused by the fact that `${{ github.ref }}` is the "fully-formed" name, e.g. `refs/remotes/pull/623/merge`, so we end up with:

```
  You are in 'detached HEAD' state. You can look around, make experimental
  changes and commit them, and you can discard any commits you make in this
  state without impacting any branches by switching back to a branch.
  
  If you want to create a new branch to retain commits you create, you may
  do so (now or later) by using -c with the switch command. Example:
  
    git switch -c <new-branch-name>
  
  Or undo this operation with:
  
    git switch -
  
  Turn off this advice by setting config variable advice.detachedHead to false
```

To work around that, we would need to:

```yaml
  - run: git switch -C ${{ github.head_ref || github.ref_name }}
```

after the checkout in order to create a local branch name.

I think that would preserve the existing behavior for actual releases while allowing branches from forks to work too.

dlech:
I do wonder if this will fail though if the branch name is the same as an existing tag name, e.g. for a release.

dlech:
In https://github.com/Bluetooth-Devices/dbus-fast/pull/621/changes, depedabot is updating some of the version comments. Perhaps instead of deleting them, we should be updating them to the correct semver tag instead of just the major version.

bdraco:
@bluetoothbot rr

fix the pr title

bdraco:
Hmm dependabot is supposed to update them. It does on other repos. Maybe a regression

eshattow:
> Hmm dependabot is supposed to update them. It does on other repos. Maybe a regression

Do we think that dependabot ignores the comment if it doesn't match a fully qualified SemVer ?

My approach is to drop the SemVer on anything the bot is currently not updating because it's currently wrong information and I lack understanding about the bots to know if bumping i.e. "v1" to "v1.2.3" would result in the bot picking it up again on the next go-around, or if it would effectively become stale again. Given that lack of understanding, I would suggest the course of action which is not going to introduce wrong information shortly after I suggest the change.

No opinion here either way though, if you suggest to just bump everything that's currently Major-only to the Major Minor Patch SemVer, I'll have a look through it and give it a go.

PR #655, opened 2026-05-15
fix(introspection): reject entity declarations to block billion-laughs

### LINKED ISSUE #642: Security: Introspection XML parsed with vulnerable stdlib ElementTree allows billion-laughs DoS from any remote service
## Problem

`Node.parse` calls `ET.fromstring(data)` directly on the response body returned by `org.freedesktop.DBus.Introspectable.Introspect`, which is then used by `BaseMessageBus.introspect` (`message_bus.py:297-305`) and `aio.MessageBus.introspect`. Python's stdlib `xml.etree.ElementTree` is explicitly documented as "not secure against maliciously constructed data" and is vulnerable to billion-laughs / quadratic-blowup internal-entity expansion attacks (the underlying `pyexpat` parser expands internal entities by default). Any DBus service the victim introspects — including unprivileged peers on the session bus, or any TCP/abstract-socket service — can return a tiny XML payload with nested entity definitions that explodes to gigabytes during parsing.

## Why This Matters

A hostile service can permanently freeze or crash a dbus-fast client (browser front-ends, GUI integration, automated discovery tools) with a ~1 KiB Introspect reply. Particularly dangerous for daemons that auto-introspect every peer that appears on the bus.

## Suggested Fix

Use `defusedxml` (already best-practice for parsing untrusted XML):
```python
from defusedxml.ElementTree import fromstring as safe_fromstring
...
element = safe_fromstring(data)
```
Add `defusedxml` to `pyproject.toml` dependencies. Alternatively, configure a custom `XMLParser` whose `expat` parser has `DefusedXmlException`-style guards (reject DOCTYPE via the `StartDoctypeDeclHandler`).

## Details

| | |
|---|---|
| **Severity** | 🟠 High |
| **Category** | dos |
| **Location** | `src/dbus_fast/introspection.py:1, 535` |
| **Effort** | ⚡ Quick fix |

---

bdraco:
## What

Switch `Node.parse` from raw `xml.etree.ElementTree.fromstring` to a custom `pyexpat` parser that rejects entity declarations and disables parameter-entity parsing. The standard D-Bus introspection `DOCTYPE` (PUBLIC, no internal subset) still parses; only entity declarations themselves are forbidden.

## Why

The stdlib `xml.etree.ElementTree` is explicitly documented as "not secure against maliciously constructed data" — `pyexpat` expands internal entities by default, so a ~1 KiB Introspect reply with nested entity definitions can explode to gigabytes during parsing. Any peer a dbus-fast client introspects — including unprivileged peers on the session bus, or any TCP/abstract-socket service — could deliver a billion-laughs / quadratic-blowup payload in response to `org.freedesktop.DBus.Introspectable.Introspect`. Particularly dangerous for daemons that auto-introspect every peer that appears on the bus.

Existing TODO at `introspection.py:1` flagged the use of `ET.fromstring` directly on untrusted data.

## How

`_parse_introspection_xml` uses `pyexpat.ParserCreate` with a `TreeBuilder` target instead of `ET.fromstring`:

- `parser.EntityDeclHandler = _reject_entity_decl` — raises `InvalidIntrospectionError` on any `<!ENTITY>` declaration, which kills both billion-laughs (nested entities) and quadratic-blowup (single large entity referenced many times). The standard D-Bus introspection DOCTYPE has no internal subset and declares no entities, so it's untouched.
- `parser.SetParamEntityParsing(XML_PARAM_ENTITY_PARSING_NEVER)` — defense-in-depth against XXE; expat already defaults to not fetching external DTDs but this makes it explicit.
- Malformed XML still raises `ET.ParseError` (re-raised from `ExpatError`) so existing callers' error-handling continues to work.

No new runtime dependency — `dbus-fast` keeps its zero-runtime-deps stance. (The issue suggested `defusedxml`; the alternative inline path it documented turned out to be cleaner here.)

## Tests

`tests/test_introspection.py`:

- `test_introspection_rejects_billion_laughs` — three nested entities, root references the outermost.
- `test_introspection_rejects_quadratic_blowup` — single 1 KiB entity referenced 64×.
- `test_introspection_rejects_xxe_external_entity` — `<!ENTITY xxe SYSTEM "file:///etc/passwd">`.
- `test_introspection_accepts_standard_doctype` — the standard D-Bus `PUBLIC` DOCTYPE still parses.
- `test_introspection_malformed_xml_raises_parse_error` — pin that malformed XML still raises `ET.ParseError`, not a silent failure.

Existing `test_introspection_from_xml_strict` / `_sloppy` (which use `tests/data/strict-introspection.xml` — itself carrying the standard D-Bus DOCTYPE) continue to pass.

closes #642


tests/test_introspection.py
```diff
@@ -145,3 +149,63 @@ def test_default_interfaces():
     # just make sure it doesn't throw
     default = intr.Node.default()
     assert type(default) is intr.Node
+
+
+def test_introspection_rejects_billion_laughs():
```
bdraco:
```suggestion
def test_introspection_rejects_billion_laughs() -> None:
```

bdraco:
@bluetoothbot review

PR #666, opened 2026-05-17
feat(service): expose message metadata via keyword-only params

### LINKED ISSUE #486: Allow access to message metadata in high-level API
**Is your feature request related to a problem? Please describe.**
There is currently no way to access message metadata (header flags and fields) in the high-level API. Some use cases require access to this (any service implementation with privileged operations using caller identity for authorization, such as via `polkit`), and others should have it to conform to D-Bus best practices (such as utilizing the "allow interactive authorization" flag).

**Describe the solution you'd like**
I'd love to see some optional keyword-only arguments for things like message headers and flags in methods and signal callbacks. Since the library already parses method signatures to get the D-Bus type signature, I think having specific checks for defined keyword-only arguments (i.e., after a `*`) that would add this metadata would be useful.

I think the most useful field is probably `SENDER`, so having a dedicated argument just for it would be useful, and given how the other fields are mostly used to determine where the message is routed in the high-level API, there may not need to be another argument to get the full list of header fields.

Basically, what I would like is to be able to write a method like this and have it get the information:

```python
@method
def Foo(self, bar: 's', baz: 'i', *, sender: str, flags: MessageFlag) -> '': ...
```

**Additional context**
Because it would only consider keyword-only args, which are not supported by D-Bus, this feature wouldn't prevent someone from having a `sender` or `flags` positional argument that _is_ exposed to D-Bus. I think that is necessary for this to be something automatic. Because of this, it might also be useful to have an argument to the `method` decorator to change the default keywords for this metadata "injection", but I don't think it's strictly necessary.

bluetoothbot:
Lets `@dbus_method` handlers receive the inbound `Message` by declaring a keyword-only `message` parameter; it stays out of the D-Bus signature.

```python
@dbus_method()
def Foo(self, bar: 's', *, message: Message) -> '':
    if not self._authorize(message.sender, message.flags):
        raise DBusError(...)
```

Read `message.sender`, `message.flags` and the rest off it for caller-identity checks (polkit), or to honour `ALLOW_INTERACTIVE_AUTHORIZATION`. Closes #486.

Per @dlech, this exposes only the whole `Message` rather than fanning out named per-field kwargs, so any future `Message` field is reachable without touching this code. The two-bus test setup is now a typed pytest fixture, and the tests assert signature exclusion through the public `introspect()` output instead of internals. Any keyword-only name other than `message` raises `ValueError` at decoration time.


dlech:
Why do we need to break down the message into components but also pass the entire message? It seems like it would be simpler and future-proof to just pass the message object.

bdraco:
@bluetoothbot rr

bdraco:
closing and turning off explore for now since I don't have time to review the autogen prs

rhendric:
I'm not, in general, a supporter of LLM-generated code, but I am using this patch in a personal project. It seems to work, and using the feature is a huge improvement over the low-level API.

I expect that most of the required labor here is design rather than code, but I would be delighted if this feature got a human review and approval.

bdraco:
The PR could be made merge ready after a few turns but it requires human time investment to do that so it needs a human champion to move this forward.

If someone wants to cleanup the design and get a PR going that it can get a review when time is available.

rhendric:
Can you offer any hints about what you'd want to see cleaned up in the design? I think the design that the LLM produced is pretty good. The only human-raised question so far is @dlech's, about whether we should expose `message` but not its properties. I find the properties intuitive; this is the high-level API, and working with a `Message` object seems more like peeking beneath that abstraction layer than being given `sender` or `flags` parameters does. But anything is better than nothing and if this is what the maintainers want, I have no problem with making that change.

I'd be happy to pitch a new PR once I have a direction.

dlech:
@bluetoothbot do option 2 from https://github.com/Bluetooth-Devices/dbus-fast/pull/666#issuecomment-4471135049

bdraco:
@bluetoothbot rr

bdraco:
@bluetoothbot rr

tests/service/test_method_metadata_kwargs.py
```diff
@@ -0,0 +1,194 @@
...
+            "unix_fds",
+            "message",
+        }
+    )
+    assert expected == METHOD_METADATA_KWARGS
+
+
+@pytest.mark.asyncio
+async def test_sender_kwarg_receives_caller_unique_name() -> None:
+    bus1 = await MessageBus().connect()
```
bdraco: The two bus pattern should be a typed
Pytest fixture

bdraco (changes requested):

bdraco:
@bluetoothbot Rr

bdraco:
@bluetoothbot rr

Fixture never made it in the push

bdraco:
pushed some changes. left it as draft. probably needs some more thought

PR #675, opened 2026-05-17
ci: cache poetry venv keyed on lockfile and cython sources

bdraco:
## What

Caches the poetry-managed `.venv` plus the built Cython artefacts (`src/dbus_fast/**/*.so`, `src/dbus_fast/**/*.c`) for both the `test` matrix and `benchmark` jobs, keyed on:

- runner OS + python version + extension mode (skip_cython / use_cython)
- `hashFiles('poetry.lock', 'pyproject.toml', 'build_ext.py')` — dependency / build set
- `hashFiles('src/dbus_fast/**/*.py', 'src/dbus_fast/**/*.pxd')` — Cython sources

On a cache hit the `Install Dependencies` step is skipped via `if: steps.cache-venv.outputs.cache-hit != 'true'`. Pure-Python and Cython source changes both invalidate the cache, so we never run tests against stale `.so` files.

`POETRY_VIRTUALENVS_IN_PROJECT: "true"` is set at the job level so poetry creates `.venv` inside the repo rather than under `~/.cache/pypoetry/virtualenvs/`, which makes it cacheable as a single path.

## Why

`actions/setup-python` with `cache: "poetry"` only caches poetry's *download* cache, not the venv. On the last green run that left `poetry install --only=main,dev` taking ~22s per matrix cell (mostly Cython compile in `use_cython` mode), running 13× across the test matrix + benchmark.

With this in place, runs where neither `poetry.lock` nor any Cythonized source has changed will reuse the prebuilt venv + `.so` files and skip the install step entirely. PRs that touch only tests, glib code, or docs will get this hit; PRs that change `src/dbus_fast/**/*.py` or `.pxd` invalidate the cython-source half of the key and rebuild as today.

## Notes

- apt caching is in #674; s390x caching is a separate follow-up.
- The cache key intentionally hashes `pyproject.toml` and `build_ext.py` so changes to `TO_CYTHONIZE`, extension flags, or runtime deps bust the cache.
- `cache: "poetry"` on `setup-python` is left in place — it still helps on the cold-cache path because poetry can install from the cached wheel store.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -37,6 +37,8 @@ jobs:
           - "skip_cython"
           - "use_cython"
     runs-on: ${{ matrix.os }}
+    env:
+      POETRY_VIRTUALENVS_IN_PROJECT: "true"
     steps:
       - uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # v6.0.2
       - name: Install libs
@@ -56,7 +58,16 @@ jobs:
           python-version: ${{ matrix.python-version }}
           cache: "poetry"
           allow-prereleases: true
+      - name: Cache poetry venv
+        id: cache-venv
+        uses: actions/cache@27d5ce7f107fe9357f9df03efb73ab90386fccae # v5.0.5
+        with:
+          path: |
+            .venv
+            src/dbus_fast/**/*.so
+          key: venv-v1-${{ runner.os }}-${{ env.ImageOS }}-${{ env.ImageVersion }}-py${{ matrix.python-version }}-${{ matrix.extension }}-${{ hashFiles('poetry.lock', 'pyproject.toml', 'build_ext.py') }}-${{ hashFiles('src/dbus_fast/**/*.py', 'src/dbus_fast/**/*.pxd') }}
       - name: Install Dependencies
+        if: steps.cache-venv.outputs.cache-hit != 'true'
         run: |
           if [ "${{ matrix.extension }}" = "skip_cython" ]; then
             SKIP_CYTHON=1 poetry install --only=main,dev
@@ -106,6 +117,8 @@ jobs:
 
   benchmark:
     runs-on: ubuntu-latest
+    env:
+      POETRY_VIRTUALENVS_IN_PROJECT: "true"
     steps:
       - uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # v6.0.2
       - name: Install libs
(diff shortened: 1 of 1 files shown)


bdraco:
Rebased onto current main + tightened the cache key for cython safety. Two concrete risks I wanted to close:

1. **Cython version drift inside the unpinned bound.** `pyproject.toml` declares `Cython>=3,<3.3.0`, so build-time isolation could resolve a different Cython between runs without changing any source file. The cached `.so` would be subtly inconsistent with what a fresh build would produce. Hashing `pyproject.toml` catches changes to the *bound* but not the *resolved version inside it*.
2. **Runner image flip.** `ubuntu-latest` periodically rolls over (22.04 → 24.04 etc.) — different glibc, different default gcc — and `runner.os` only resolves to `Linux`, so a cached `.so` built against the old image could be served against the new one.

Changes on this revision:

- Added `${{ env.ImageOS }}` and `${{ env.ImageVersion }}` to the cache key — these are provided by GitHub-hosted runners and tick whenever the image rolls. Forces a rebuild on image flip.
- Added a `v1` prefix to the key as an explicit bust knob if we ever need to invalidate everything by hand.
- Dropped `src/dbus_fast/**/*.c` from the cached paths. On a cache hit we skip `Install Dependencies` entirely, so build_ext never re-reads the `.c` files; only the `.so` matters at runtime. Cuts cache size.

Failure mode if a bad cached `.so` somehow gets through: pytest fails at collection-time on `import dbus_fast`, which is loud and easy to diagnose — no silent corruption. The Cython-drift concern (#1) is still theoretical; if it bites, the `v1` prefix is the kill switch.

bdraco:
Correction to my earlier safety comment: I made too much of the Cython-drift risk. Walking through it more carefully —

A cache hit requires both source AND `poetry.lock` to be unchanged. Whenever `poetry lock` resolves a new Cython, the lockfile hash changes and the cache invalidates. So a hit always serves a `.so` built with the same Cython that the current lockfile pins. The only thing the cache *doesn't* protect against is two *fresh* builds picking different Cython versions out of the build-system range — that's a reproducible-builds concern, not a cache-correctness concern.

Closed #678 (the build-system pin) for the same reason: it'd restrict downstream sdist builds without buying us anything the existing key doesn't already give us. The `v1` prefix on the cache key is still the manual bust knob if anything ever does go sideways.

The other safety bits on this revision (`ImageOS`/`ImageVersion` in the key, dropping `*.c` from cached paths) still apply.

PR #676, opened 2026-05-17
ci: cache apt and uv downloads for the s390x big-endian job

bdraco:
## What

Adds the first caching layer to the `test_big_endian` (s390x) job, which previously had **zero** caching:

1. `actions/cache` for `~/cache-s390x/apt` (apt `.deb` archives) and `~/cache-s390x/uv` (uv download cache), keyed on `pyproject.toml` + `poetry.lock` with a `s390x-deps-v1-` prefix so the bust knob is explicit.
2. `dockerRunArgs` bind-mounts those host paths into the QEMU container at `/var/cache/apt/archives` and `/root/.cache/uv` respectively.
3. Inside the container, drops the default `docker-clean` apt config and writes `Binary::apt::APT::Keep-Downloaded-Packages "true";` so the `.deb` files survive `apt-get install`.
4. After the action exits, a `sudo chown -R` on `~/cache-s390x` reclaims the mount contents from root (the container writes as root via the bind mount) so `actions/cache` can save them.

## Why

The s390x leg is the slowest job in CI by a large margin — about 9 minutes on the latest green run (#25999269859) vs ~90s for any other test cell — because everything inside the QEMU emulator was uncached: `apt-get update && install`, `pip install uv`, every dependency wheel.

Caching the apt deb archives + the uv wheel cache cuts the QEMU-side reinstall cost on every push where `poetry.lock` / `pyproject.toml` haven't changed. The Cython compile under emulation still happens (that's the larger remaining cost) but those minutes only need to be amortized across the wheel-download cost going forward; caching the `.so` files inside the inner `git clone` is fragile under the action's bind-mount layout, so it's left for a follow-up.

## Notes

- The `s390x-deps-v1-` prefix in the key lets us bust the cache deliberately (e.g. if apt archive layout changes) by bumping the integer.
- apt deb caching for the host-side `test` and `benchmark` jobs is in #674; venv caching is in #675.
- The `restore-keys` fallback (`s390x-deps-v1-`) means a lockfile bump still warms from the previous cache and only fetches new wheels/debs rather than the full set.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -89,13 +89,31 @@ jobs:
     steps:
       - name: Checkout code
         uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # v6.0.2
+      - name: Cache s390x apt + uv downloads
+        uses: actions/cache@27d5ce7f107fe9357f9df03efb73ab90386fccae # v5.0.5
+        with:
+          path: |
+            ~/cache-s390x/apt
+            ~/cache-s390x/uv
+          key: s390x-deps-v1-${{ hashFiles('pyproject.toml', 'poetry.lock') }}
+          restore-keys: |
+            s390x-deps-v1-
+      - name: Prepare s390x cache mountpoints
+        run: mkdir -p ~/cache-s390x/apt ~/cache-s390x/uv
       - uses: uraimo/run-on-arch-action@1c358dc49363439f8c563ce8f93005f7fe76b849 # v3
         name: Run commands
         id: runcmd
         with:
           arch: s390x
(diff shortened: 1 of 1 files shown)


bdraco:
Tightened the reclaim step: `sudo chown -R` → `sudo chown -h -R` (`--no-dereference`).

Without `-h`, chown follows symlinks and modifies ownership of their *targets*. The reclaim runs against directories populated by code inside the QEMU container (apt's deb cache and uv's wheel cache), so any symlink that ever made it into those dirs would let the container influence ownership of arbitrary host paths. Unlikely in practice with apt/uv, but `-h` closes the door for ~zero cost — chown applies to the symlink itself instead of dereferencing.

PR #678, opened 2026-05-17
build: pin cython to 3.2.4 for reproducible builds

bdraco:
## What

Pins Cython to an exact version (`3.2.4`, the version that was already resolving in `poetry.lock`) in two places:

- `[tool.poetry.group.dev.dependencies]` — was `>=3,<3.3.0`
- `[build-system].requires` — was `>=3,<3.3.0`

## Why

Both spots previously declared a range. `[build-system].requires` is the more important one: it's resolved fresh by `pip` build-isolation on every install, *not* from `poetry.lock`, so two builds on the same source tree could pick different Cython versions whenever a new patch landed within the range. Different Cython versions emit subtly different `.c` (and therefore `.so`), which:

- defeats reproducible wheel builds, and
- breaks any cache keyed on source hashes (see #675), because the cached `.so` could disagree with what a fresh build would produce even though no source file changed.

Pinning makes the build deterministic — the same source tree always produces the same `.c`/`.so`. Bumping Cython is now an explicit PR that runs CI + CodSpeed, instead of an invisible drift.

## Notes

- `poetry.lock` already resolved Cython to `3.2.4`, so this just removes the freedom poetry/pip had to drift inside the range. No runtime resolution change.
- The unrelated `urllib3` group-list adjustment in `poetry.lock` is a side-effect of re-running `poetry lock` with poetry 2.2.1 — it corrects a stale group assignment that was already in the lock; it doesn't change which packages are actually installed in either group.
- Future bumps should be normal PRs; dependabot can still propose them.

### DIFF
--- a/poetry.lock
+++ b/poetry.lock
@@ -1272,7 +1272,7 @@ version = "2.7.0"
 description = "HTTP library with thread-safe connection pooling, file post, and more."
 optional = false
 python-versions = ">=3.10"
-groups = ["dev", "docs"]
+groups = ["docs"]
 files = [
     {file = "urllib3-2.7.0-py3-none-any.whl", hash = "sha256:9fb4c81ebbb1ce9531cce37674bbc6f1360472bc18ca9a553ede278ef7276897"},
     {file = "urllib3-2.7.0.tar.gz", hash = "sha256:231e0ec3b63ceb14667c67be60f2f2c40a518cb38b03af60abc813da26505f4c"},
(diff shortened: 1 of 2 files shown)


bdraco:
Closing without merging — on reflection, the build-system pin would restrict downstream sdist builds without actually being required for the cache safety argument in #675.

The cache key already hashes `poetry.lock`, so any Cython version change (whenever someone runs `poetry lock`) invalidates the cache. The only thing the build-system pin would buy is bit-for-bit reproducible wheels across fresh builds, which is a separate goal we don't claim here. The `v1` prefix on the cache key is still there as a manual bust knob if a problematic Cython release ever does sneak through.

PR #691, opened 2026-05-20
fix(aio): send hello via sock_sendall to keep connect off the event loop

bdraco:
The hello message in `MessageBus.connect()` was written via `self._stream.write(serialized); self._stream.flush()`, which Home Assistant Supervisor flagged via blockbuster as a synchronous `io.BufferedWriter.write` running on the event loop (see comment thread on home-assistant/supervisor#6853).

Switched it to `await self._loop.sock_sendall(self._sock, serialized)`, matching the pattern already used by `_authenticate()` on the same socket. As a side effect, almost all of the entries in `tests/conftest.py::_KNOWN_BLOCKING` now pass cleanly, so popped them; two test-side `socket.sendall` calls in `test_auth_readline.py` remain (peer-side setup, not a production bug) and stay xfailed for a follow-up.


bdraco:
@bluetoothbot review

PR #692, opened 2026-05-20
test: drop blocking sendall in auth_readline peer-side setup

bdraco:
The two `tests/test_auth_readline.py` cases set up their peer socket with a blocking `socket.sendall`, which blockbuster catches even though it is test-only and not a production bug. Switched both peers to non-blocking and drove the writes via `loop.sock_sendall`; the 64 KiB junk-line test runs sendall as a background task so the write progresses concurrently with the reader and does not deadlock when the socketpair send buffer is small.

Cleared `_KNOWN_BLOCKING` and dropped the `PytestUnraisableExceptionWarning` and `ResourceWarning` ignores from filterwarnings; no xfails remain.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -68,12 +68,6 @@ addopts = "-v --cov=dbus_fast --cov-report=term-missing:skip-covered"
 pythonpath = ["src"]
 filterwarnings = [
     "error",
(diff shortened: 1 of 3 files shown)


bdraco:
@bluetoothbot review

bdraco:
big endian test failure is unrelated infra failure

PR #697, opened 2026-05-20
fix(introspection): bound <node> nesting depth to prevent DoS

### LINKED ISSUE #696: Security: unbounded recursion in Node.from_xml — introspection XML DoS via nested <node> elements
## Problem

`Node.from_xml` recurses into every child `<node>` element with no depth bound: `node.nodes.append(Node.from_xml(child, ...))`. The existing hardening at `_parse_introspection_xml` closes DTD/entity-based amplification (billion laughs, XXE) but does nothing about deeply nested element trees, which expat happily builds and ElementTree happily iterates. A peer that controls the response to `org.freedesktop.DBus.Introspectable.Introspect` can return an XML document of the form `<node><node><node>…</node></node></node>` whose depth exceeds Python's recursion limit (~1000 frames ≈ ~250 nested `<node>` elements at ~4 frames per `Node.from_xml`). The `RecursionError` raised in `Node.parse` propagates out of `BaseMessageBus.introspect`'s `reply_notify` and is delivered to the caller as an introspection failure, but on Python ≤ 3.11 a very deep recursion can also exhaust the C stack and crash the interpreter (the recursion limit is checked but the per-frame C stack consumption is not strictly bounded). Same path applies to the `<node>` parsing inside `xml.etree`'s tree-build step, but that one is bounded by expat's own depth limit; the second-stage Python recursion in `from_xml` is the unbounded one.

## Why This Matters

A malicious or buggy service on the bus that the client introspects (directly via `MessageBus.introspect()`, or indirectly because the high-level client constructs proxies that introspect) returns a poisoned XML payload. Under default Python settings the client-side coroutine that awaited `introspect()` raises `RecursionError`, which most callers treat as an unexpected failure and either crash the integration path or retry forever. Less severe than the unmarshaller variant DoS because the connection is not torn down, but it gives any name-owning peer a deterministic way to break clients that introspect them.

## Suggested Fix

Bound `<node>` nesting in `Node.from_xml` (object paths have a natural depth — D-Bus paths are typically ≤10 components deep; a cap of 32 covers any plausible real path with margin). Add a `_depth` keyword parameter that defaults to 0, increment on each recursive call, and raise `InvalidIntrospectionError` if it exceeds the cap. Alternatively, iterate the element tree to a flat list of (parent, child) pairs and build the `Node` tree non-recursively. The recursive shape is fine; only the missing bound is the bug.

## Details

| | |
|---|---|
| **Severity** | 🟡 Medium |
| **Category** | dos |
| **Location** | `src/dbus_fast/introspection.py:531-565 (Node.from_xml), 567-589 (Node.parse)` |
| **Effort** | ⚡ Quick fix |

---

bdraco:
`Node.from_xml` recursed into each child `<node>` with no depth bound, so a peer that controls an Introspect reply could send `<node><node>...</node></node>` deep enough to blow the stack on the client. Real object paths are only a handful of components deep; this caps recursion at 32 and raises `InvalidIntrospectionError` past that.

Fixes #696

### DIFF
--- a/src/dbus_fast/introspection.py
+++ b/src/dbus_fast/introspection.py
@@ -532,10 +532,7 @@ class Node:
 
     @staticmethod
     def from_xml(
-        element: ET.Element,
-        is_root: bool = False,
-        validate_property_names: bool = True,
-        _depth: int = 0,
+        element: ET.Element, is_root: bool = False, validate_property_names: bool = True
     ) -> "Node":
         """Convert an :class:`xml.etree.ElementTree.Element` to a :class:`Node`.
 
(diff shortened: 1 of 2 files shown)


bdraco:
@bluetoothbot review

PR #698, opened 2026-05-20
fix(unmarshaller): cap container nesting depth to prevent stack overflow DoS

### LINKED ISSUE #695: [Security] high finding — details withheld (PVRS unavailable)
A security finding was identified during an automated audit but could not be submitted via Private Vulnerability Reporting (PVRS).

**Severity**: high
**Category**: dos

Details have been withheld from this public issue to prevent disclosure of exploitable vulnerabilities. Please review the audit logs or contact the security team for full details.

---

bdraco:
A peer can pack roughly forty million levels of variant-in-variant into a single 128 MiB message, since each level only costs three bytes on the wire (signature_len byte, `v`, null). `_read_variant` and `read_array` recursed into each level and could blow the Python C stack before the existing size cap rejected the message.

The fix adds a single cdef unsigned int depth counter to the `Unmarshaller`, bumped at entry to each container reader (variant, array, struct, dict entry) and decremented before the return, after any recursive call has popped its own frame. If it exceeds 64 (the same limit the D-Bus spec already implies for any single signature, 32 array plus 32 struct), parsing raises `InvalidMessageError` and the connection tears down before the stack overflows.

Verified the Cython HTML for `unmarshaller.html`; the counter ops are score-0 pure C (`__pyx_v_self->_container_depth = (... + 1);` etc.) and the `result is None` check in the refactored `_read_variant` compiles to a pointer compare. The only Python-level code on the new path is the f-string in the reject branch.

Closes #695

### DIFF
--- a/src/dbus_fast/_private/unmarshaller.pxd
+++ b/src/dbus_fast/_private/unmarshaller.pxd
@@ -24,6 +24,7 @@ cdef unsigned int LITTLE_ENDIAN
 cdef unsigned int BIG_ENDIAN
 cdef unsigned int PROTOCOL_VERSION
 cdef unsigned int _MAX_MESSAGE_SIZE
+cdef unsigned int _MAX_CONTAINER_DEPTH
 
 
 cdef unsigned int HEADER_PATH_IDX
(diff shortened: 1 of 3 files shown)


bdraco:
@bluetoothbot review

